PR #11, opened 2014-10-30
Add cpu-microcode detect-plugin.

xnox:
Whitelist microcode packages for autoinstallation.

Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/ubuntu-drivers-common/+bug/1386257

Signed-off-by: Dimitri John Ledkov <dimitri.j.ledkov@intel.com>

### DIFF
--- a/detect-plugins/cpu-microcode.py
+++ b/detect-plugins/cpu-microcode.py
@@ -24,8 +24,7 @@ def detect(apt_cache):
             for line in file:
                 if line.startswith('vendor_id'):
                     cpu = line.split(':')[1].strip()
-                    return [db.get(cpu)]
+                    if cpu in db:
+                        return [db.get(cpu)]
     except IOError as err:
         logging.debug('could not open /proc/cpuinfo: %s', err)
-
-    return None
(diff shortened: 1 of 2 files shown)


xnox:
May I upload this into ubuntu?

detect-plugins/cpu-microcode.py
```diff
@@ -0,0 +1,31 @@
...
+    'AuthenticAMD': 'amd64-microcode',
+     }
+
+def detect(apt_cache):
+    try:
+        with open('/proc/cpuinfo') as file:
+            for line in file:
+                if line.startswith('vendor_id'):
+                    cpu = line.split(':')[1].strip()
+                    return [db.get(cpu)]
```
martinpitt: If the vendor is not in db, then this will return [None]. Would be better to do something like

     try:
         return [db[cpu]]
     except KeyError:
         break

?

martinpitt:
Aside from that, it looks good to me, thanks!

detect-plugins/cpu-microcode.py
xnox: Yes that would be better. Also other plugins do not have return, or return None if nothing is detected. So that would work. Thanks for this. I'll land this now, to get into feature freeze =/ =)

martinpitt:
Thanks! Merged.

PR #15, opened 2015-12-29
Depend on liblocale-gettext-perl for autopkgtest

cjwatson:
This forces it not to be dropped during Perl transitions (thereby breaking
the fglrx test) due to pinning only necessary packages from -proposed.

This corresponds to an upload I just made to xenial.


martinpitt:
Thanks!

PR #18, opened 2016-05-18
Quirks/quirkinfo.py: Handle UnicodeDecodeError

grandpaul:
On some Dell machine, the /sys/class/dmi/id/board_name is ffffffffffff0a (hex). It means empty.
This makes an error when reading these strings in python3. We can
catch UnicodeDecodeError and ignore this kind of strings.

Signed-off-by: Ying-Chun Liu (PaulLiu) <paulliu@debian.org>


tseliot:
Merged. Thanks for your contribution!

PR #19, opened 2016-12-08
Handle AMDGPU-Pro switchable graphics

khfeng:


tseliot:
Is there a reason to default to power saving mode?

khfeng:
I guess when users want to switch from discrete only to hybrid graphics, they want to use integrated GPU. But this is just my guess, either powersaving or performance should be fine.

PR #21, opened 2017-06-16
gpu-manager: Disable 'nvidiaXineramaInfo' for NVIDIA X driver

fourdollars:
On X servers with RandR 1.2, disabling "nvidiaXineramaInfo" causes the
NVIDIA X driver to still register its Xinerama implementation but report
a single screen-sized region.

### DIFF
--- a/share/hybrid/gpu-manager.c
+++ b/share/hybrid/gpu-manager.c
@@ -2050,6 +2050,7 @@ static bool write_prime_xorg_conf(struct device **devices, int cards_n) {
                 "    Device \"nvidia\"\n"
                 "    Option \"AllowEmptyInitialConfiguration\" \"on\"\n"
                 "    Option \"IgnoreDisplayDevices\" \"CRT\"\n"
+                "    Option \"nvidiaXineramaInfo\" \"False\"\n"
                 "EndSection\n\n",
                (int)(devices[i]->bus),
                (int)(devices[i]->domain),


tseliot:
ubuntu-drivers-common 1:0.4.23 (in Ubuntu 17.10) supports using custom xorg.confs in the /usr/share/gpu-manager.d directory. On non hybrid systems, you have to name the file "non-hybrid", whereas, if it's a hybrid system, you can use the "hybrid-performance" file and the "hybrid-power-saving" file for performance mode and power saving mode, respectively.

fourdollars:
The default value of nvidiaXineramaInfo is still true because of the backward compatibility concern from NVIDIA.
However the concern doesn't seem to be needed for Ubuntu anymore because all of https://launchpad.net/ubuntu/+source/libxrandr are greater than 1.2.
Would it be better to just disable it directly for Ubuntu by default?

tseliot:
While it's true that we have relied on RandR 1.2 for years, I am very conservative when it comes to introducing default driver options. Can you point me to any bug reports that are affecting by this feature, please?

fourdollars:
LP: #1693717 is a private bug. I don't know if you can see it.
If you can, please leave some comment on it.

tseliot:
I wrote a comment there. Changing a default option may cause regressions, especially for use cases we might not be familiar with (e.g. using X remotely with different RandR protocols, etc.). This is why I developed the custom xorg.conf feature in gpu-manager.

PR #24, opened 2018-01-26
Remove the cpu-microcode.py detection plugin

chrisccoulson:
Kernel metapackages are going to have a hard dependency on the
{intel,amd64}-microcode packages to ensure that all bare-metal installs
benefit from microcode updates. This means that microcode packages are
no longer "additional drivers" and since they won't be able to be removed
without uninstalling the kernel metapackages, there is no need for them
to be exposed by ubuntu-drivers.

See http://bugs.launchpad.net/bugs/1738259.


tseliot:
You changes look good to me.

Thanks for your contribution!

PR #29, opened 2019-03-11
Whitelist open-vm-tools*

jibel:
Added open-vm-tools* to the whitelist to auto-install the driver on
VMware guests


tseliot:
Thanks!

PR #30, opened 2019-03-14
Free only drivers

jibel:
This branch adds a --free-only option to ubuntu-drivers to only consider free drivers.

The fake archive of the test suite had to be modified to support free (main and universe) and non-free components. 
Tests have been added to cover this option.

The function _is_package_free has been fixed.


tseliot:
Thanks

PR #32, opened 2019-04-25
Remove obsolete open-vm-dkms detect plugin

jibel:
open-vm-dkms has been removed from the archive and detection of VMware guest is now done with modalias. This plugin is obsolete so removed it.


tseliot:
Merged. Thanks

PR #33, opened 2019-04-30
Fix nvidia detector

jibel:
Make nvidia-detector work again.

Fixed regexp to match all sub classes of display devices.
Actually print the name of the selected driver.
Removed python2 code since it runs with python3.
Commented out unused code.
Enabled static tests.


tseliot:
Thanks!

PR #35, opened 2019-08-18
gpu-manager.c: Turn on PRIME offload when in on-demand mode

philipl:
I believe the intention of the on-demand PRIME mode is to support
selectively using the nvidia GPU on a per-window basis with the new
offload feature. Assuming that's true, we need to add the special
ServerLayout option to actually enable it.


share/hybrid/gpu-manager.c
```diff
@@ -1410,6 +1410,42 @@ static bool create_prime_outputclass(void) {
...
+static bool create_offload_serverlayout(void) {
+    _cleanup_fclose_ FILE *file = NULL;
+    _cleanup_free_ char *multiarch = NULL;
+    char command[100];
+    char xorg_d_custom[PATH_MAX];
+
+    snprintf(xorg_d_custom, sizeof(xorg_d_custom), "%s/11-nvidia-offload.conf",
+             xorg_conf_d_path);
+
+    snprintf(command, sizeof(command),
```
tseliot: Why do you need multiarch to create the file?

philipl: Sorry, I left that in accidentally when I copied the outputclass function. I'll remove it.

tseliot:
Merged. Thanks!

PR #36, opened 2019-08-27
gpu-manager: set power to "auto" in PRIME "ONDEMAND" mode; minor cleanups

dadap:
We could possibly go further and have power set to "auto" for PRIME "ON" mode as well, but I didn't do that in this patch series since I wasn't sure if there was a particular reason it's set to "on" in that mode. (e.g. to avoid an interaction problem with other power management tools like bbswitch)


tseliot:
It's all merged now, except for b356d2e792d5d69647f9d7928681c51de1a5fd5f. Thanks

PR #41, opened 2020-02-12
gpu-manager: delete offload configuration file on NVIDIA single-GPU systems

dadap:
On systems with only an NVIDIA GPU present, the render offload configuration file generated by gpu-manager when "on-demand" mode is enabled prevents X from starting, as it explicitly configures the NVIDIA GPU to drive a GPU screen with a non-NVIDIA protocol screen. This could happen if a muxed hybrid system was prevoiusly configured in hybrid mode with render offloading and then reconfigured in dGPU-only mode, or if a boot disk is migrated from a hybrid system to a non-hybrid NVIDIA-only system.

In order to allow X to continue working when on-demand mode is configured on an NVIDIA-only system, delete the 11-nvidia-offload.conf file if it exists. The on-demand configuration in /etc/prime-discrete is left in place, allowing 11-nvidia-offload.conf to be generated again if booting to a hybrid graphics system again.

### DIFF
--- a/share/hybrid/gpu-manager.c
+++ b/share/hybrid/gpu-manager.c
@@ -2271,7 +2271,9 @@ int main(int argc, char *argv[]) {
             }
         }
         else if (boot_vga_vendor_id == NVIDIA) {
-            fprintf(log_handle, "Nothing to do\n");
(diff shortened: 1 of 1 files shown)


tseliot:
Merged. Thanks!

PR #43, opened 2020-04-22
find_reverse_deps: Use exact matching with both installed and candidate

xnox:
On Ubuntu livecd, we may have packages that are installed, but have
new candidates. Or those that are installed and cannot be installed
again.

Consider a case where and old iso boots into
linux-image-5.4.0-25-generic, yet linux-image-5.4.0-26-generic is
available. There is internet, and apt update was executed. In such
situation, linux-meta-package discovery was failing, because there
were no longer any candidates that depend on the old 25 abi, and only
metas that depend on 26.

Also, I switched from vague matching across an unparsed raw Depends
string, to iterating the parsed depedencies instead, such that
matching is exact. This prevents false positive find in case we have
non-unique suffixes, i.e. linux-modules-nvidia-440-lowlatency &
linux-modules-nvidia-440-lowlatency-hwe-20.04. A reverse-dep search
for one, must not find the other one.

I fixed this in both funcations that try to hunt for
reverse-dependencies. Maybe the two functions can be unified, with
like kwarg filter= to allow additional filtering as required in the
kerneldetect case vs detect case.

With this changes in place I can see that ubuntu-drivers list
correctly resolves for modules to be provided by l-r-m meta-packages,
instead of dkms.

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -907,18 +907,22 @@ def find_reverse_dependencies(apt_cache, package, prefix):
     '''Return the reverse dependencies for a package'''
     # prefix to restrict the searching
     # package we want reverse dependencies for
-    deps = []
+    deps = set()
     for pkg in apt_cache:
-        if (pkg.name.startswith(prefix)):
-            try:
-                dependencies = apt_cache[pkg.name].candidate.\
-                         record['Depends']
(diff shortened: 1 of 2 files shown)


mwhudson:
This makes sense to me fwiw

tseliot:
It looks good to me. Thanks

PR #44, opened 2020-05-01
gpu-manager.c: remove dependency of dpkg-architecture from package dp…

ycheng:
call uname instead of dpkg-architecture.

fix bug LP: #1875339


tseliot:
Hi, thanks for your work. It think the correct solution is to add the missing dependency to ubuntu-drivers-common, though. uname is not always reliable, and I don't want tests to break in chroots.

ycheng:
IMHO, install dpkg-dev not seems right for me. I don't have much knowledge on uname, maybe something else more light-weighted shall just work.

PR #45, opened 2020-07-30
checking supported-gpus.json to see whether the gpu is runtime pm supported

cyruslien:
gpu-manager.c: checking supported-gpus.json to see whether the gpu is runtime pm supported
debian/control: depend on jq


tseliot:
Thanks for your contribution. ubuntu-drivers-common can't depend on a packages that is in universe. Packages that are available in main, can only depend on packages in main.

PR #49, opened 2021-01-19
debian/rules: make tests on riscv64 non-fatal.

xnox:
powerpc architecture last featured in xenial. riscv64 is new
architecture in focal, where some tests are failing. In later
releases, unittests are disable by policy (aka builds are done as
nocheck). Thus ensure that package builds on focal, when backported.


xnox:
https://bugs.launchpad.net/ubuntu/focal/+source/ubuntu-drivers-common/+bug/1912202

PR #51, opened 2021-03-31
drop dependency on alsa-utils

mwhudson:
We want to start using ubuntu-drivers in the server installer but do not
want alsa-utils there. The code using aplay to detect certain kinds of
modem will fail to detect them if aplay is not installed but will not
crash. As alsa-utils is in the desktop-common seed there should be no
loss of functionality anywhere ubuntu-drivers is used today.


tseliot:
Ok, merged. Thanks.

PR #57, opened 2021-09-14
SRU for regression in the installer

vorlonofportland:


sil2100 (approved):
Ok, this feels like something we should have regardless if this is the best fix for the underlying issue. In general the code looks legit - let me give it a spin and see if it fixes the issue for me.

I would still be interested what actually causes LP: #1943614 in bionic.

tseliot:
I wasn't aware of the problem. The workaround looks fine to me, thanks

PR #60, opened 2021-10-08
Impish ubiquity

tseliot:
This allows enabling KMS, and picking up the configuration which ubuntu-drivers is supposed to generate for the target chroot in ubiquity.

### DIFF
--- a/ubuntu-drivers
+++ b/ubuntu-drivers
@@ -198,20 +198,17 @@ def command_install(args):
     if with_nvidia_kms:
         UbuntuDrivers.detect.set_nvidia_kms(1)
 
-    # --package-list should not install any packages
-    if args.package_list:
-        with open(args.package_list, 'a') as f:
-            f.write('\n'.join(to_install))
-            f.write('\n')
-            return 0
-
     ret = subprocess.call(['apt-get', 'install', '-o',
         'DPkg::options::=--force-confnew', '-y'] + to_install)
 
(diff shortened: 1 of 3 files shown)


ubiquity/target-config/31ubuntu_driver_packages
```diff
@@ -7,7 +7,19 @@ set -e
...
     apt-install $p
 done
 
+# Call prime-select, if needed
+if [ -e /run/nvidia_runtimepm_supported ]; then
+    apt-install nvidia-prime
+    cp /run/nvidia_runtimepm_supported /target/run/nvidia_runtimepm_supported
+    rm -f /target/etc/prime-discrete
+    chroot /target prime-select on-demand </dev/null
+fi
```
xnox, line 25: please have new line at the end of file.

ubuntu-drivers
```diff
@@ -184,21 +185,33 @@ def command_install(args):
...
+                version = int(package.split('-')[-1])
+            except:
+                pass
+            finally:
+                with_nvidia_kms = version >= 470
+
+    if with_nvidia_kms:
         UbuntuDrivers.detect.set_nvidia_kms(1)
 
+    # --package-list should not install any packages
```
xnox, line 201: This will break networking & storage drivers, as at the moment ubiquity does expect that things do get installed in the live session when called with package-list I think. Since it should be installing things like wifi drivers, ethernet drivers, storage drivers if any.

I think it might make more sense to like skip installing nvidia things, cause well, installing nvidia in live session is never nice.

PR #63, opened 2021-11-30
Optimize packages_for_modalias

mwhudson:
By compiling all the Modaliases into one big regexp that can be used to
quickly reject most devices.


mwhudson:
I haven't tested this on a machine where `ubuntu-drivers list` should return something yet, but on my laptop it takes runtime from about 4.5s to 1.5s and given that `python3 -c 'import apt; list(apt.Cache())'` takes ~1s I'm not sure there's a heap more to save (unless we can find a way to filter to the packages with Modaliases fields faster)

tseliot:
It does lower runtime here too. I'm merging this. Thanks

PR #67, opened 2022-02-18
Make sure nvidia-drm rules action processed

os369510:
This MP is a better solution for LP#1958488.

Verified passed on problematic machine in 1000 times stress.

GDM expects nvidia-drm be loaded to configures wayland or xorg:
```
KERNEL!="nvidia_drm", GOTO="gdm_nvidia_drm_end"
SUBSYSTEM!="module", GOTO="gdm_nvidia_drm_end"
ACTION!="add", GOTO="gdm_nvidia_drm_end"
ATTR{parameters/modeset}!="Y", GOTO="gdm_disable_wayland"
```
In some race condition, the nvidia-drm was load later than gdm.
gpu-manager can cover this case to prevent race condition.


khfeng:
This method works because `71-u-d-c-gpu-detection.rules` is executed after `61-gdm.rules`. So I think it's a good idea to mention _why_ it works in the commit message and comment.

Otherwise LGTM.

tseliot (approved):
It all looks good to me. Thanks

PR #68, opened 2022-03-04
Drop apt in favour of the apt_pkg bindings

tseliot:
This leads to visible performance gains.

The test suite has been updated accordingly.

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -81,14 +81,12 @@ def _check_video_abi_compat(apt_cache, package):
             logging.debug('Cannot find %s package in the cache. Cannot check ABI' % (xorg_driver_name))
             return True
 
-    records = apt_pkg.PackageRecords(apt_cache)
     depcache = apt_pkg.DepCache(apt_cache)
     candidate = depcache.get_candidate_ver(package)
-    records.lookup(candidate.file_list[0])
-    deps = records['Depends'].split(', ')
+
     needs_video_abi = False
-    for dep in deps:
-        if dep.startswith('xorg-video-abi-'):
+    for dep_name, dep_ver, dep_op in candidate.depends_list_str.get('Depends'):
+        if dep_name.target_pkg.name.startswith('xorg-video-abi-'):
             needs_video_abi = True
             break
 
@@ -104,11 +102,10 @@ def _check_video_abi_compat(apt_cache, package):
         return True
 
     candidate = depcache.get_candidate_ver(xorg_core)
-    records.lookup(candidate.file_list[0])
 
-    provides = records.__getitem__('Provides').split(', ')
-    for provide in provides:
-        if provide.startswith('xorg-video-abi-'):
+
+    for provides_name, provides_ver, p_version in candidate.provides_list:
+        if provides_name.startswith('xorg-video-abi-'):
             xorg_video_abi = provide
 
     if xorg_video_abi:
@@ -1222,7 +1219,6 @@ def find_reverse_dependencies(apt_cache, package, prefix):
 
 def get_linux_image_from_meta(apt_cache, pkg):
     depcache = apt_pkg.DepCache(apt_cache)
-    records = apt_pkg.PackageRecords(apt_cache)
 
     try:
         candidate = depcache.get_candidate_ver(apt_cache[pkg])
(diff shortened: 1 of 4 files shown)


UbuntuDrivers/detect.py
```diff
@@ -70,58 +70,91 @@ def system_modaliases(sys_path=None):
...
+        xorg_core = apt_cache.__getitem__('xserver-xorg-core')
     except KeyError:
+        logging.debug('xserver-xorg-core not available, cannot check ABI')
         return True
-    if 'xorg-video-abi-' in deps and xorg_video_abi not in deps:
+
+    candidate = depcache.get_candidate_ver(xorg_core)
+    records.lookup(candidate.file_list[0])
+
+    provides = records.__getitem__('Provides').split(', ')
```
julian-klode, line 109: No need for records, Use candidate.provides_list instead, that's more correct and much faster.

UbuntuDrivers/detect.py
```diff
@@ -70,58 +70,91 @@ def system_modaliases(sys_path=None):
...
-            if p.startswith('xorg-video-abi-'):
-                xorg_video_abi = p
-                # logging.debug('_check_video_abi_compat(): Current X.org video abi: %s', xorg_video_abi)
-                break
-    except (AttributeError, KeyError):
-        logging.debug('_check_video_abi_compat(): xserver-xorg-core not available, cannot check ABI')
+    if package.name.startswith('nvidia-driver-'):
+        xorg_driver_name = package.name.replace('nvidia-driver-', 'xserver-xorg-video-nvidia-')
+        try:
+            package = apt_cache.__getitem__(xorg_driver_name)
```
julian-klode, line 79: Why write it this way instead of `apt_cache[xorg_driver_name]`, same in other places?

UbuntuDrivers/detect.py
```diff
@@ -70,58 +70,91 @@ def system_modaliases(sys_path=None):
...
+        except KeyError:
+            logging.debug('Cannot find %s package in the cache. Cannot check ABI' % (xorg_driver_name))
+            return True
+
+    records = apt_pkg.PackageRecords(apt_cache)
+    depcache = apt_pkg.DepCache(apt_cache)
+    candidate = depcache.get_candidate_ver(package)
+    records.lookup(candidate.file_list[0])
+    deps = records.__getitem__('Depends').split(', ')
```
julian-klode, line 88: Use `candidate.depends_list` or `candidate.depends_list_str` and lose the record parsing.

UbuntuDrivers/detect.py
```diff
@@ -171,67 +203,96 @@ def packages_for_modalias(apt_cache, modalias):
...
 
-    for o in pkg.candidate.origins:
-        if o.origin == 'Ubuntu':
-            return True
-    return False
+    records = apt_pkg.PackageRecords(apt_cache)
+    records.lookup(candidate.file_list[0])
+
+    try:
+        origin = records.__getitem__('Origin')
```
julian-klode, line 251: Consider using `candidate.file_list_list[0][0].origin` instead of doing the record lookup. I don't know why the field is copied to Packages files in the first place in Ubuntu.

UbuntuDrivers/detect.py
```diff
@@ -464,8 +530,10 @@ def _get_headless_no_dkms_metapackage(pkg, apt_cache):
         package = apt_cache.__getitem__(candidate)
         # skip foreign architectures, we usually only want native
         # driver packages
-        if (package.candidate and
-                package.candidate.architecture in ('all', system_architecture)):
+        package_candidate = depcache.get_candidate_ver(package)
+        records.lookup(package_candidate.file_list[0])
```
julian-klode, line 534: The records are never read, so this seems wasteful to lookup

UbuntuDrivers/kerneldetection.py
```diff
@@ -173,19 +185,25 @@ def get_linux_version(self):
         linux_image_meta = self.get_linux_image_metapackage()
         linux_version = ''
         try:
-            dependencies = self.apt_cache[linux_image_meta].candidate.\
-                             record['Depends']
+            # dependencies = self.apt_cache[linux_image_meta].candidate.\
+            #                  record['Depends']
+            records = apt_pkg.PackageRecords(self.apt_cache)
+            candidate = self.apt_depcache.get_candidate_ver(self.apt_cache[linux_image_meta])
+            records.lookup(candidate.file_list[0])
```
julian-klode, line 192: You don't seem to be using the records, get rid of the initialization and the lookup

julian-klode (changes requested):
Performance review mostly.

Generally confused by why you call `a.__getitem__(b)` like that instead of using `a[b]`.

PR #72, opened 2022-06-21
Skip to wait nvidia-drm for nvidia-340

os369510:
nvidia-340 doesn't have nvidia-drm module.
Thus, it will unconditionally wait 10 seconds during booting (LP: #1973604)


os369510:
Here is the test PPA https://launchpad.net/~os369510/+archive/ubuntu/lp1973604 which is based on `focal` branch and confirmed by user.

tseliot:
Merged, thanks

PR #73, opened 2022-06-21
Skip to wait nvidia-drm for nvidia-340

os369510:
nvidia-340 doesn't have nvidia-drm module.
Thus, it will unconditionally wait 10 seconds during booting (LP: #1973604)


tseliot:
Merged, thanks

PR #71, opened 2022-04-11
Insert a hook to check the allowing list to force install nvidia

os369510:
This is a first version and waiting for verification.
It works good in `ubuntu-drivers debug` but I need to have a build on different systems.

Tried three scenario on a focal I+N laptop.

1. the force install list contains a older version
```
    {
      "devid": "0x24BA",
      "name": "TEST 24BA",
      "minimumbranch": "470",
      "features": [
        "runtimepm"
      ]
    }
```
the output
```
$ ubuntu-drivers debug
...
Found a specific nv driver version 470 for TEST 24BA(0x24BA)
...
=== matching driver packages ===
nvidia-driver-470: installed: <none>   available: 470.103.01-0ubuntu2  [distro]  non-free  modalias: pci:v000010DEd000024BAsv0000103Csd000089C6bc03sc00i00  path: /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0  vendor: NVIDIA Corporation  support level: LTSB 
nvidia-driver-510: installed: 510.54-0ubuntu0.20.04.1   available: 510.60.02-0ubuntu1 (auto-install)  [distro]  non-free  modalias: pci:v000010DEd000024BAsv0000103Csd000089C6bc03sc00i00  path: /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0  vendor: NVIDIA Corporation  support level: PB
```

2. the force install list contains the same version from ubuntu archive
```
    {
      "devid": "0x24BA",
      "name": "TEST 24BA",
      "minimumbranch": "510",
      "features": [
        "runtimepm"
      ]
    }
```
```
$ ubuntu-drivers debug
...
Found a specific nv driver version 510 for TEST 24BA(0x24BA)
...
=== matching driver packages ===
nvidia-driver-510: installed: 510.54-0ubuntu0.20.04.1   available: 510.60.02-0ubuntu1 (auto-install)  [distro]  non-free  modalias: pci:v000010DEd000024BAsv0000103Csd000089C6bc03sc00i00  path: /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0  vendor: NVIDIA Corporation  support level: PB 
```

3. the force install list contains the a non-exist version.
```
    {
      "devid": "0x24BA",
      "name": "TEST 24BA",
      "minimumbranch": "580.1234",
      "features": [
        "runtimepm"
      ]
    }
```
```
ubuntu-drivers debug
...
[...]

PR #76, opened 2022-09-16
Os369510 force install nvidia cleanup

tseliot:
This is a slightly cleaned up version of pull request #71 by @os369510.

Here are the main changes:

1) grouped up the commits into unified commits, and cleaned up the addition and subsequent removal of the changelog entry;

2) Changed the word "customized" with "custom". For example, the json file is now /etc/custom_supported_gpus.json

3) Fine tuned the descriptions in the documentation.


os369510:
@tseliot yes, it looks good to me. thanks a lot

PR #77, opened 2022-10-05
add some error trapping when aplay is not installed (on servers)

bladernr:
Adds error trapping to clean up traceback when aplay is not installed.  This is preferable to adding a dependency on aplay as this would install unnecessary other dependency packages on servers.  This just clears up the traceback with a helpful message to remind the user that aplay just isn't there.


tseliot:
Hi, thanks for the patch. I have just pushed the main commit (0848f433076e9878fcacf8d5bba601d127006dee), which is now in master. Thanks again.

PR #79, opened 2022-11-06
fix: local variable 'version' referenced before assignment:

2976560783:
when ValueError happened, the code in finally part will raise UnboundLocalError: local variable 'version' referenced before assignment. So move with_nvidia_kms before except part.


tseliot:
Merged. Thank you for your contribution.

PR #80, opened 2022-12-06
removing redundant variable

Alex-S-H-P:
### Changes done

**nvidia_desktop_pre_installation_hook** contained a `with_nvidia_kms` boolean variable that could be eliminated by some minor refactoring. As only the set of this variable was affected by the inner for loop, we can also `return` as soon as a true is found.

### Rationale

I had an issue with the variable "version" being read without being set, thus raising an `UnboundLocalError`. I fixed it on my machine (and fixed my arguments, since that was what showed the issue), and looked online. It seems to be fixed here, but my local fix seemed slightly better. If it is not, please just delete this and message me telling me what I overlooked.


tseliot (approved):
Approved. Thank you for you contribution.

PR #90, opened 2023-09-28
Add boradcom-sta-dkms to allowlist.

xnox:
In Lunar it was identified that both bcmwl and broadcom-sta-dkms are identical and maintained in the archive twice. They were merged, taking the Debian name as the new default one. However, allowlist was forgotten to be updated in ubuntu-drivers to actually offer default install of broadcom-sta-dkms.

LP: #2013236


tseliot:
It makes sense. Merged. Thanks.

os369510:
Do we need `get_userspace_lrm_meta()` for something like dkms cases when secure boot enabled?

xnox:
@os369510 this driver is not signed, and never will be. Installer at runtime offers to autogenerate MOK key and enroll MOK key instead on the secureboot systems.

PR #91, opened 2023-11-10
Support the installation of MIPI camera out of tree drivers

xanthein:
Support to install MIPI camera out of tree modules including

- linux-modules-ipu6*
- linux-modules-ivsc*
- linux-modules-usbio*

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -24,6 +24,26 @@ system_architecture = ''
 lookup_cache = {}
 custom_supported_gpus_json = '/etc/custom_supported_gpus.json'
 
+class MIPICamPkgNameInfo(object):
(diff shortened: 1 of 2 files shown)


tseliot:
What are the names of the linux-modules-ipu6*, linux-modules-ivsc*, linux-modules-usbio* supposed to look like? And why is there no kernel flavour detection?

Also, can you provide a test for the code in  in tests/test_ubuntu_drivers.py please?

os369510:
+@xnox for awareness.

I believe you are reviewing the modalias design of MIPI device drivers and there is a plan for put them in extra-modules.
Would you please share your opinions?

xnox:
@tseliot we need something like this, plus probably have archive provided stems, and the whole lot.

PR #98, opened 2024-07-04
share/hybrid: Change Makefile to respect CC and CFLAGS environment variables

kukrimate:
During the introduction of the new frame pointer policy in Ubuntu, it was noticed that the produced gpu-manager binary does not have frame pointer.

It seems like this is due to the Makefile hard-coding the compile flags.

This patch changes the Makefile to use CC and CFLAGS from the environment (and also simplifies it when implicit rules already do the right thing).

### DIFF
--- a/share/hybrid/Makefile
+++ b/share/hybrid/Makefile
@@ -1,19 +1,14 @@
 #!/usr/bin/make -f
 
-PROGRAM = gpu-manager
-PROGRAM_FILES = gpu-manager.c
-CC = gcc
-CFLAGS =-g -Wall $(shell pkg-config --cflags libpci libdrm libkmod) -Ijson-parser json-parser/json.c
-LDFLAGS =$(shell pkg-config --libs libpci libdrm libkmod) -lm
+PROG = gpu-manager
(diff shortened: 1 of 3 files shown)


share/hybrid/Makefile
```diff
@@ -1,19 +1,14 @@
...
-PROGRAM_FILES = gpu-manager.c
-CC = gcc
-CFLAGS =-g -Wall $(shell pkg-config --cflags libpci libdrm libkmod) -Ijson-parser json-parser/json.c
-LDFLAGS =$(shell pkg-config --libs libpci libdrm libkmod) -lm
+PROG = gpu-manager
+OBJ = gpu-manager.o json-parser/json.o
 
-all: build
+CFLAGS += -Wall $(shell pkg-config --cflags libpci libdrm libkmod)
+LDLIBS += $(shell pkg-config --libs libpci libdrm libkmod) -lm
```
gagath, line 7: For both of these lines I think the [override directive](https://www.gnu.org/software/make/manual/html_node/Override-Directive.html) should be used.

gagath (changes requested):

PR #99, opened 2024-07-05
Recommend hwe-*-meta for non-oem hardware enablement meta packages

tobhe:
The plan is to use this namespace for harware enablement meta packages that are not part of the official certification process. They must be recommended to have subiquity pick them up on installation.


julian-klode:
👍 ship it

PR #100, opened 2024-07-15
Wait for nvidia-drm to settle before opening /dev/dri/card*

aleasto:
Move the udev settling logic earlier, before trying to open /dev/dri/card* because doing so too early may trigger a nvidia bug: https://forums.developer.nvidia.com/t/545-29-06-18-1-flip-event-timeout-error-on-startup-shutdown-and-sometimes-suspend-wayland-unusable/274788

With this fix in place we can enable nvidia-drm.fbdev=1 by default, fixing bug LP: #2060268

### DIFF
--- a/share/hybrid/gpu-manager.c
+++ b/share/hybrid/gpu-manager.c
@@ -2988,6 +2988,34 @@ int main(int argc, char *argv[]) {
         pci_cleanup(pacc);
         pacc = NULL;
     }
+
+    /* Probe graphic drivers earlier since some graphic related application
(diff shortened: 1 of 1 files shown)


vanvugt:
Is this still needed given the new proposed plan [in comments 61/62](https://bugs.launchpad.net/bugs/2060268)?

aleasto:
The new default configuration will work without this, but if the user manually enables `nvidia-drm.fbdev=1`  they will get a broken graphical session without this.

Also `nvidia-drm.fbdev=1` might become enabled by default in the future, although one would hope that the issue this is working around will be properly resolved by nvidia by then.

PR #101, opened 2024-07-16
Remove SimpleDRM device when nvidia-drm loads

aleasto:
This normally happens automatically for DRM devices that also register a framebuffer device, but that's not the case yet for the nvidia driver.

Newer nvidia drivers (>=545) are now providing a framebuffer device, but it's disabled by default and we still support older drivers.

(LP: #2060268)


aleasto:
Crucially this keeps the simple-framebuffer device around so that VTs keep working.

PR #102, opened 2024-07-31
jammy LP: #2060268

aleasto:
jammy is going to get kernel 6.8-hwe which includes SimpleDRM and thus lp#2060268


aleasto:
`Skip to wait nvidia-drm for nvidia-340` was backported to focal but not to jammy. I've included it here since it fixes a merge conflict, and i think it makes sense to have it.

PR #104, opened 2024-09-19
Improve SimpleDRM+NVIDIA fix across kernel versions

aleasto:
Commit 9eac534db0013aff9b9124985dab114600df9081 in kernel 6.11.0 has changed the sysfs path of simple-framebuffer.0 which is now a child of some PCI device.

Use udev to keep track of the simple-framebuffer device regardless of its parent device. If by the time the nvidia drm device has registered SimpleDRM is still around (always the case in single-GPU systems without nvidia-drm.fbdev=1), remove it.

LP: #2060268

### DIFF
--- a/share/hybrid/71-u-d-c-gpu-detection.rules
+++ b/share/hybrid/71-u-d-c-gpu-detection.rules
@@ -3,7 +3,9 @@
 # Remove SimpleDRM device when nvidia-drm loads.
 # This normally happens automatically for DRM devices that also register
 # a framebuffer device, but that's not the case yet for the nvidia driver.
-ACTION=="add", SUBSYSTEM=="module", KERNEL=="nvidia_drm", TEST=="/sys/devices/platform/simple-framebuffer.0/drm/card0", RUN+="/bin/rm /dev/dri/card0"
+ACTION=="add", KERNEL=="card0", SUBSYSTEM=="drm", DRIVERS=="simple-framebuffer", ENV{U_D_C_IS_SIMPLEDRM}="1", RUN+="/bin/touch /run/u-d-c-card0-is-simpledrm"
+ACTION=="remove", KERNEL=="card0", SUBSYSTEM=="drm", ENV{U_D_C_IS_SIMPLEDRM}=="1", RUN+="/bin/rm -f /run/u-d-c-card0-is-simpledrm"
+ACTION=="add", KERNEL=="card[0-9]*", SUBSYSTEM=="drm", DRIVERS=="nvidia", TEST=="/run/u-d-c-card0-is-simpledrm", RUN+="/bin/rm /dev/dri/card0"
 
 # Create a file with the card details for gpu-manager
 ACTION=="add", SUBSYSTEM=="drm", DEVPATH=="*/drm/card*", RUN+="/sbin/u-d-c-print-pci-ids"


aleasto:
I'd also like a review from @vanvugt 

Daniel, did you originally use the event for _module-loaded_ over using the event for _drm-device-registered_ for a specific reason?
For this change i had to switch to use the latter because the former raced with the _simpledrm-added_ event, in oracular only interestingly. I consistently receive events in the following order:
1. nvidia-drm module loads
2. simpledrm registers card0
3. nvidia registers card1

share/hybrid/71-u-d-c-gpu-detection.rules
```diff
@@ -3,7 +3,9 @@
 # Remove SimpleDRM device when nvidia-drm loads.
 # This normally happens automatically for DRM devices that also register
 # a framebuffer device, but that's not the case yet for the nvidia driver.
-ACTION=="add", SUBSYSTEM=="module", KERNEL=="nvidia_drm", TEST=="/sys/devices/platform/simple-framebuffer.0/drm/card0", RUN+="/bin/rm /dev/dri/card0"
+ACTION=="add", KERNEL=="card0", SUBSYSTEM=="drm", DRIVERS=="simple-framebuffer", ENV{U_D_C_IS_SIMPLEDRM}="1", RUN+="/bin/touch /run/u-d-c-card0-is-simpledrm"
+ACTION=="remove", KERNEL=="card0", SUBSYSTEM=="drm", ENV{U_D_C_IS_SIMPLEDRM}=="1", RUN+="/bin/rm -f /run/u-d-c-card0-is-simpledrm"
```
vanvugt, line 7: `ACTION=="remove"`... doesn't sound testable right now. I imagine you would need a different kernel config where simpledrm is a module to test that. But for completeness it probably makes sense.

vanvugt (approved):
Overall this sounds right. But we'll need more people actually testing it (sorry I don't have time today).

vanvugt:
> Daniel, did you originally use the event for _module-loaded_ over using the event for _drm-device-registered_ for a specific reason?

No good reason - I didn't know enough about the syntax and once I had a single line rule that solved the bug reliably in testing I had no reason to search for alternatives.

> For this change i had to switch to use the latter because the former raced with the _simpledrm-added_ event, in oracular only interestingly. I consistently receive events in the following order:
> 
> 1. nvidia-drm module loads
> 2. simpledrm registers card0
> 3. nvidia-drm registers card1

Extra robustness is fine of course, but that ordering sounds suspicious. SimpleDRM should be loaded in-kernel (`grep CONFIG_DRM_SIMPLEDRM /boot/config-*`) so I am very surprised if `nvidia-drm` is loaded before card0 was registered. Or maybe not surprised... DRM device startup is historically very slow (up to 10 seconds) which has caused Plymouth problems for years.

share/hybrid/71-u-d-c-gpu-detection.rules
aleasto: This is not the event for module loaded, but for device add/remove, so it is testable.
I actually expected _remove_ to trigger every time i915 or amdgpu loads removing `/dev/dri/card0`, but apparently the whole sequence normally happens fast enough that udev isn't notified of card0 add nor remove.

However I tested it simulating an extra-slow module load time:
* append `modprobe.blacklist=amdgpu` to the kernel cmdline
* boot to a tty
* verify `/run/u-d-c-card0-is-simpledrm` exists
* `modprobe amdgpu`
* verify `/run/u-d-c-card0-is-simpledrm` was removed

The remove part is important because on multi-gpu systems after SimpleDRM is removed the card0 name would then be available for other GPUs to register and we need to make sure we're not removing those.

vanvugt: It sounds like you're right, I'm just nervous about adding complexity. At least it will be confined to 24.10 until it's had time to mature before any backports...

This needs testing by at least one more person. If not me in the next few days then anyone else who can run Nvidia (ideally testing both single GPU and hybrid GPUs).

aleasto: > ideally testing both single GPU and hybrid GPUs

Right, hybrids is not something I can test

vanvugt:
Verified on:
* Nvidia single GPU ([LP#2060268](https://bugs.launchpad.net/bugs/2060268) is fixed again)
* Intel single GPU (no change)
* Intel+Nvidia hybrid (no change)

luisalvarado:
Oh my, took me long to find this, thank you @vanvugt was about to report the ghost 46' monitor I do not have. Was almost ALMOST ready to start searching for a weird cable connected to the PC that was long enough to hide a 46' monitor somewhere.

luisalvarado:
Oh my, took me long to find this, thank you @vanvugt was about to report the ghost 46' monitor I do not have. Was almost ALMOST ready to start searching for a weird cable connected to the PC that was long enough to hide a 46' monitor somewhere.

This is only happening on Wayland just in case. I cant see the dual ghost on X11. This also could be related to the whole Nvidia Settings app being mostly empty on Wayland, but showing properly on X11 for the 555 and 560 drivers.

PR #113, opened 2025-01-24
Implement kernel version check at start of list or install

MitchellAugustin:
To ensure users are not running an outdated prebuilt kernel (which could cause compatibility issues if the recommended Nvidia drivers are only built for a more recent kernel version), and to ensure that the user is aware of the implications of the DKMS module usage if running a custom kernel for which DKMS modules would be required, add the following behavior to the "install" and "list" commands (and their variants):

- If the user is running a kernel version that is not provided by an Ubuntu kernel package, require them to pass --include-dkms to proceed (detected by matching `uname -r` output against Ubuntu version format, then checking to ensure Ubuntu kernel metapackage is installed)
- If the user is running an Ubuntu kernel, and any installed Ubuntu kernel metapackages have updates available according to the apt cache, warn them to upgrade before proceeding.

It is worth discussing whether we should eventually be more assertive about having the Ubuntu kernel up-to-date, but for now, it is only implemented as a warning.

### DIFF
--- a/UbuntuDrivers/kerneldetection.py
+++ b/UbuntuDrivers/kerneldetection.py
@@ -25,6 +25,8 @@ import logging
 import re
 
 from subprocess import Popen
+import os
+from apt_pkg import version_compare
 
 
 class KernelDetection(object):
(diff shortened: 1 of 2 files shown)


MitchellAugustin:
I tested the DKMS error functionality by verifying that the DKMS warning prints on my machine which is running a custom kernel. I tested the kernel version functionality by adding a PPA with a newer kernel version of the same upgrade path to my machine, then verifying that the kernel version warning prints when my machine's kernel is still on the original version.
I observed no regressions when doing a control test on a machine with an up-to-date Ubuntu kernel.

PR #116, opened 2025-04-01
Performance fix and improvement

schopin-pro:
Here are a couple of commits to first fix a [performance regression](https://bugs.launchpad.net/ubuntu/+source/ubuntu-drivers-common/+bug/2092601) and then improve upon it with judicious use of caching. More details in the individual commit logs.

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -207,7 +207,7 @@ def _check_video_abi_compat(apt_cache, package):
     return True
 
 
-def _apt_cache_modalias_map(apt_cache):
+def apt_cache_modalias_map(apt_cache):
     '''Build a modalias map from an apt_pkg.Cache object.
 
     This filters out uninstallable video drivers (i. e. which depend on a video
@@ -294,21 +294,17 @@ def package_get_nv_allowing_driver(did):
     return version
 
 
-def packages_for_modalias(apt_cache, modalias):
+def packages_for_modalias(apt_cache, modalias, modalias_map=None):
     '''Search packages which match the given modalias.
 
     Return a list of apt.Package objects.
     '''
     pkgs = set()
 
-    apt_cache_hash = hash(package.get_fullname() for package in apt_cache.packages)
-    try:
-        cache_map = packages_for_modalias.cache_maps[apt_cache_hash]
-    except KeyError:
-        cache_map = _apt_cache_modalias_map(apt_cache)
-        packages_for_modalias.cache_maps[apt_cache_hash] = cache_map
+    if modalias_map is None:
+        modalias_map = apt_cache_modalias_map(apt_cache)
 
-    pat, bus_map = cache_map.get(modalias.split(':', 1)[0], (None, {}))
+    pat, bus_map = modalias_map.get(modalias.split(':', 1)[0], (None, {}))
     vid, did = _get_vendor_model_from_alias(modalias)
     nvamd = None
     found = 0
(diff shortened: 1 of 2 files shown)


bdrung (approved):
This change looks good to me.

schopin-pro:
Looking a bit more into the history of the code, there is no specified reason for hashing on the package names. Until the migration to `apt_pkg.Cache` in 8b24cbaec878969f685f9325513a6a8f9275ba6a, it was using the hash of `apt_cache` itself. It does introduce a nice property of being able to cache the modalias map even if the cache object has changed as long as its contents stay the same, but was it really intended in the first place?

If not, a simple `apt_cache_hash = hash(apt_cache)` would have done the job just as well.

ktpawlak:
unfortunately fixing cache so it performs as before, breaks one test:

======================================================================
FAIL: test_system_driver_packages_force_install_nvidia (test_ubuntu_drivers.DetectTest.test_system_driver_packages_force_install_nvidia)
system_driver_packages() force install config points to an older version.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.13/unittest/mock.py", line 1426, in patched
    return func(*newargs, **newkeywargs)
  File "/<<PKGBUILDDIR>>/tests/test_ubuntu_drivers.py", line 1468, in test_system_driver_packages_force_install_nvidia
    self.assertFalse('nvidia-driver-510' in res_wrong_json)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: True is not false

======================================================================
FAIL: test_pycodestyle_clean (test_static.TestStatic.test_pycodestyle_clean)
pycodestyle - Python style guide checker
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/tests/test_static.py", line 95, in test_pycodestyle_clean
    self.assertEqual(0, len(output))
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
AssertionError: 0 != 2

I tried to fix caching in a different way, but encountered the same thing and had to drop it

schopin-pro:
To be clear, are those test failures happening with the PR code or the simpler fix I suggested in my comment?

ktpawlak:
this is with changes from this PR.

schopin-pro:
I spent way too much time thinking of how to detect when a apt cache has changed, so that we can have a proper key for our hash. That was the wrong way of looking at the problem.

As usual, stepping back to consider the broader picture was the right thing to do. The problems are:

* Modalias map computation is really expensive. That's the motivation behind attempting to caching it.
* That computation depends on the state of the APT cache, we don't control its lifecycle and it doesn't provide easy ways to tell if it has changed.

The thing is, the second problem is only a problem because we want to do local caching, and we need to do cache invalidation. The performance problems are because we're just *very* bad at it.

So another solution is not to use a cache at all but just to let the caller provide a modalias map of their own. SInce the caller presumably has more knowledge of the apt_cache lifecycle, they can reuse the map if need be (and boy do they need to).

Now, I still haven't figured out how to properly run the tests, but I do think that my new approach has the merits of being simpler and shouldn't break anything :)

schopin-pro:
The relevant tests seem to pass.

```bash
❯ PYTHONPATH=. tests/run tests.test_ubuntu_drivers
/home/schopin/dev/ubuntu-drivers-common/tests/test_ubuntu_drivers.py:21: PyGIWarning: UMockdev was imported without specifying a version first. Use gi.require_version('UMockdev', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import UMockdev
[skip]
----------------------------------------------------------------------
Ran 53 tests in 250.317s

OK (skipped=1)
```

ktpawlak:
one backport coming right up

PR #118, opened 2025-04-22
Provide a sample output for system_device_drivers.

nteodosio:
For systems without additional drivers available, provide an option to create a mock result.

Intended for facilitating debugging for the developer of consumer applications. For example, I did this to test software-properties, for https://launchpad.net/bugs/2107741 and some bugs before it which I do not recall. As another developer also mentioned no return value for the function in his system, I'm proposing this mock output.


ktpawlak:
sorry, but test code needed to simulate bugs in other software won't be merged.

PR #126, opened 2025-07-09
Implement kernel version and secure boot warnings and fix test cases

MitchellAugustin:
DKMS module installation will require MOK enrollment on systems with Secure Boot. This patch makes users aware of the subsequent steps required for driver functionality if they are on an impacted deployment.

Test failures were introduced in b5f42cd4216ec420373ec823f1414c7e4b772d6f since the mock environment doesn't use sudo. This adds an alternate path for mock invocation of the tool from such an environment.


MitchellAugustin:
Removed sudo test case fixes since Kuba addressed these upstream. Rebuilt successful in this PPA: https://launchpad.net/~mitchellaugustin/+archive/ubuntu/ubuntu-drivers-kernel-check-jul9-hopefullyfinal/

MitchellAugustin:
@mwhudson @julian-klode would one of you (or anyone in Foundations) like to review this PR?

julian-klode (approved):
I'm unhappy so much is duplicated here and I have a question about printing in the diff, but otherwise this lgtm

PR #133, opened 2025-08-19
Add new ubuntu-drivers welcome page

MitchellAugustin:
This commit adds a new welcome page, shown when ubuntu-drivers is run without subsequent options, which displays the following:
- The kernel and NVIDIA modules that will be booted next
- Current NVIDIA kernel module status
- Information on any unexpected kernel/driver mismatches or bad states, with suggestions for how to proceed (reboot, rerun u-d-c, etc)
- List of packages installed on the system that were likely installed by u-d-c


MitchellAugustin:
Fixed & ready for review again

ktpawlak:
there is a conflict

UbuntuDrivers/detect.py
```diff
@@ -1007,6 +1008,26 @@ def system_device_drivers(apt_cache=None, sys_path=None, freeonly=False):
     return result
 
 
+def get_installed_packages_by_glob(apt_cache, glob_pattern):
```
mwhudson: This could be implemented via apt-patterns like f"apt list '(?and(?installed,?name({fnmatch.translate(glob_pattern(})))" but it looks like the relevant bits of the API are not exposed to Python...

UbuntuDrivers/detect.py
```diff
@@ -1698,3 +1719,261 @@ def get_linux_modules_metapackage(apt_cache, candidate):
         pass
 
     return metapackage
+
+
+def _get_kernel_from_grub_entry(entry_num):
```
mwhudson: I can't say I'm a big fan of parsing grub config like this but I'm not sure what a better idea is. If you structure this so the parsing of the config is separated from the loading of it, at least you can write some tests!

UbuntuDrivers/detect.py
```diff
@@ -1698,3 +1718,333 @@ def get_linux_modules_metapackage(apt_cache, candidate):
...
+    Args:
+        grub_cfg_path: Path to grub.cfg file (default: /boot/grub/grub.cfg)
+
+    Returns:
+        String content of grub.cfg, or None if file doesn't exist or can't be read
+    '''
+    try:
+        if not os.path.exists(grub_cfg_path):
+            logging.debug('Could not read %s in _load_grub_cfg()', grub_cfg_path)
+            return None
```
mwhudson: I don't see the point of this 'look before you leap' style but whatever :-)

UbuntuDrivers/detect.py
```diff
@@ -1698,3 +1718,333 @@ def get_linux_modules_metapackage(apt_cache, candidate):
...
+            universal_newlines=True,
+            stderr=subprocess.DEVNULL
+        )
+
+        # Get the first match if any
+        matches = result.strip().split('\n')
+        if matches and matches[0]:
+            return matches[0]
+
+    except (subprocess.CalledProcessError, FileNotFoundError) as e:
```
mwhudson: I don't think we need to worry about 'find' not being present, that would indicate a _severely_ broken system!

PR #136, opened 2025-09-01
Only list the target driver for custom configuration

KaiChuan-Hsieh:
Respect the custom configuration if the NV driver series is specified. Fix runtimepm feature obtain if -open is specified in the branch field.

The selecting driver algorithm will prioritize LTSB over PB and NFB even though a custom configuration was set like series 570 or 575. The ubuntu-drivers will only list the series specified in the custom_supported_gpu.json according to  user's configuration.

Ex:
On machine with Nvidia dGPU device id [10de:2db8]

custom_supported_gpus.json
{
    "chips": [
       {
          "devid": "0x2DB8",
          "name": "NVIDIA RTX PRO 1000 Blackwell Generation Laptop GPU",
          "branch": "570-open",
          "features": [
            "dpycbcr420",
            "dpgsynccompatible",
            "hdmi4k60rgb444",
            "hdmigsynccompatible",
            "runtimepm",
            "vdpaufeaturesetK",
            "kernelopen"
          ]
      }
    ]
}

Without the change, it will list all supported drivers, and the driver with higher priority will be installed.
$ ubuntu-drivers list
nvidia-driver-580-open, (kernel modules provided by linux-modules-nvidia-580-open-oem-24.04c)
nvidia-driver-570-open, (kernel modules provided by linux-modules-nvidia-570-open-oem-24.04c)
nvidia-driver-580, (kernel modules provided by linux-modules-nvidia-580-oem-24.04c)
nvidia-driver-570, (kernel modules provided by linux-modules-nvidia-570-oem-24.04c)

With the change, it will list the target driver only, and user will install specified driver. 
$ ubuntu-drivers list
nvidia-driver-570-open, (kernel modules provided by linux-modules-nvidia-570-open-oem-24.04c)

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -319,9 +319,12 @@ def packages_for_modalias(apt_cache, modalias, modalias_map=None):
         if nvamd is not None and not found:
             logging.debug('%s is not in the package pool.' % nvamdn)
 
-    if not found:
-        if pat is None or not pat.match(modalias):
-            return []
+    if found:
+        logging.info('Found custom %s in the package pool.' % nvamdn)
+        return [apt_cache[nvamdn]]
+
+    if pat is None or not pat.match(modalias):
+        return []
 
     for alias in bus_map:
         if fnmatch.fnmatchcase(modalias.lower(), alias.lower()):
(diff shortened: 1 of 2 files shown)


MitchellAugustin:
@KaiChuan-Hsieh Can you give an example reproducer for the bug you see, as well as the expected output (with your fix) + the actual output (without your fix)?

KaiChuan-Hsieh:
> @KaiChuan-Hsieh Can you give an example reproducer for the bug you see, as well as the expected output (with your fix) + the actual output (without your fix)?

@MitchellAugustin I add the example and impact in the first comment, please help to check, thanks.

MitchellAugustin (changes requested):
I left one inline comment on a comparator that I think we should adjust a bit.

binli:
Currently 595 PB is by default on 24.04 and 26.04 even 580 is LTSB.
With this patch, I could select the 580 as the default driver.

```
$ cat /etc/custom_supported_gpus.json 
{
  "chips": [
    {
      "devid": "0x27B0",
      "name": "NVIDIA RTX 4000",
      "branch": "580.173",
      "features": [
        "runtimepm"
      ]
    }
  ]
```
```
$ ubuntu-drivers list --recommended
nvidia-driver-580 linux-modules-nvidia-580-generic-hwe-24.04
$ sudo mv /etc/custom_supported_gpus.json .
$ ubuntu-drivers list --recommended
nvidia-driver-595-open linux-modules-nvidia-595-open-generic-hwe-24.04
```

PR #138, opened 2025-09-18
Prevent attempts to coinstall conflicting nvidia-driver-* packages (LP: #2125156)

MitchellAugustin:
Attempting to run `sudo ubuntu-drivers install` or `sudo ubuntu-drivers install --gpgpu` on some machines that support multiple Nvidia driver version branches can sometimes result in the generated `apt install` command including multiple conflicting packages, which causes the installation to be stopped.

Since only one package matching the `nvidia-driver-*` glob pattern should be installable in a valid solution, adjust the filtering logic to remove any `nvidia-driver-*` recommendations after the first.

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -1032,6 +1032,12 @@ def get_desktop_package_list(
 
             candidate = depcache.get_candidate_ver(package_obj)
             records.lookup(candidate.file_list[0])
+
+            # Do not add more than one nvidia-driver-* (or associated packages) to to_install
+            if p.startswith("nvidia-driver-"):
+                if any(pkg.startswith("nvidia-driver-") for pkg in to_install):
+                    continue
+
             to_install.append(p)
 
             # See if runtimepm is supported
--- a/ubuntu-drivers
+++ b/ubuntu-drivers
@@ -267,6 +267,11 @@ def install_gpgpu(args):
                        key=cmp_to_key(lambda left, right:
                                       UbuntuDrivers.detect._cmp_gfx_alternatives_gpgpu(left[0], right[0])),
                        reverse=True):
+        # Do not add more than one nvidia-driver-* (or associated packages) to to_install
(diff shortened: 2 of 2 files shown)


UbuntuDrivers/detect.py
```diff
@@ -1007,6 +1007,34 @@ def system_device_drivers(apt_cache=None, sys_path=None, freeonly=False):
...
+    Args:
+        package_list: A list of package name strings.
+
+    Returns:
+        A new list with the specified items filtered out.
+    '''
+    filtered_list = []
+    found_first_nvidia_driver = False
+
+    for package in package_list:
```
ogandojose, line 1026: @MitchellAugustin is this list already sorted by number (i.e. 580...570...550....) ?

ogandojose (commented):
Asking context questions in the code.

UbuntuDrivers/detect.py
MitchellAugustin: @ogandojose actually, to be more precise, I should say it does not deviate from the original sorting logic defined elsewhere in the code - it will just remove subsequent numbered versions after the first one from the to_install list (since they would conflict).

In my experience this usually results in the highest supported version number being picked. (Regardless, this patch does not change the ordering of the numbered versions with respect to each other in the final install list).

ktpawlak:
my thoughts would be to catch this error from happening, so that the function that generates a list of packages to install stops after one matching set, instead of post filtering the list.

MitchellAugustin:
@ktpawlak I revised this as suggested. I also confirmed that it [passes all tests](https://launchpad.net/~mitchellaugustin/+archive/ubuntu/udc-resolver-bug-9-15-25/+packages) and confirmed that the resolver issues are no longer present on my original container reproducer, and that I'm not seeing any regressions on 8 other machines I have access to with NV GPUs.
[Logs attached, in case you want to confirm that the package install lists look as expected](https://github.com/user-attachments/files/22601826/udc-resolver-fix-logs.txt)

PR #144, opened 2026-01-08
Consolidate GPGPU apt filtering functionality into detect.py

MitchellAugustin:
Prior to this change, the gpgpu driver installation procedure
involves additional filtering of package candidates in
./ubuntu-drivers, which the test cases (which are testing for
 end-user package filtering expectations) and any other direct
callers of detect.py's filter functions are unaware of.

This change moves all filtering logic to detect.py and adjusts
all associated test cases as needed to represent the actual view
of which drivers the end-user will have installed.

Direct invocation of auto install filter also includes
nvidia-driver-XXX in tests now

This is a corrected version of 4fd9ff3d6102392b6985ff891db989eed39010ef, which forward-ports the coinstallation bug fix in 29f21023a31692a284a56772e349cbd6dd0c578f

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -1015,7 +1015,6 @@ def get_desktop_package_list(
         apt_cache, sys_path, freeonly=free_only,
         include_oem=include_oem)
 
-    to_install = []
     to_install = auto_install_filter(apt_cache, include_dkms, packages, driver_string, get_recommended=False)
     if not to_install:
         logging.debug('No drivers found for installation.')
@@ -1094,6 +1093,8 @@ def _sort_packages_by_preference(packages, gpgpu=False):
 
     Args:
         packages: Dict of package candidates to sort.
+            Typically originates from system_driver_packages() or system_gpgpu_driver_packages(),
+            and is filtered depending on the mode (desktop vs gpgpu) and user preferences.
         gpgpu: Boolean flag indicating whether to use GPGPU (server) sorting preferences.
 
     Returns:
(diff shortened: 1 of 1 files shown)


MitchellAugustin:
Prior review: https://github.com/canonical/ubuntu-drivers-common/pull/128

debian/changelog
```diff
@@ -1,9 +1,14 @@
-ubuntu-drivers-common (1:0.10.3ubuntu1) plucky; urgency=medium
+ubuntu-drivers-common (1:0.10.4) resolute; urgency=medium
```
mwhudson: er what's going on here

UbuntuDrivers/detect.py
```diff
@@ -1136,11 +1088,110 @@ def _process_driver_string(string):
     return driver
 
 
-def gpgpu_install_filter(packages, drivers_str, get_recommended=True):
+def already_installed_filter(cache, packages, include_dkms, comparator):
```
mwhudson: can you document comparator please? Is there any way in the current code where comparator is _not_ `_cmp_gfx_alternatives`?

mwhudson (commented):
OK so early next week was a lie, but anyway. Seems like a move in the right direction but I'm not quite sure I grasp it all yet. One question/request for now.

debian/changelog
MitchellAugustin: Thanks, yeah I'll just drop the changelog from this commit entirely since we're probably going to bundle it with all the other recent changes in a separate commit anyway.

UbuntuDrivers/detect.py
mwhudson: I think the thing I find strangest about all this is the `if comparator == _cmp_gfx_alternatives:` code. Comparing a comparator by value just rubs me the wrong way. Would it perhaps be cleaner to just have a "gpgpu" boolean flag?

UbuntuDrivers/detect.py
```diff
@@ -1165,11 +1088,130 @@ def _process_driver_string(string):
...
+    return to_install
+
+
+def gpgpu_install_filter(cache, include_dkms, packages, drivers_str, get_recommended=True, comparator=None):
     drivers = []
     allow = []
     result = {}
-    '''Filter the Ubuntu packages according to the parameters the users passed
+    '''
+    Filter the Ubuntu packages according to the parameters the users passed.
```
mwhudson: Again, filter *and sort*, right?

PR #156, opened 2026-02-27
Add dbus interface

jibel:
A D-Bus service is added to query ubuntu-drivers from anywhere including from a confioned package.
The service exposes a single method called `drivers`. This method returns a dict of devices and for each device the list of supported drivers with details about the driver.
The service is d-bus activated and shuts down after 5 minutes.

### DIFF
--- a/UbuntuDrivers/service/drivers_service.py
+++ b/UbuntuDrivers/service/drivers_service.py
@@ -49,9 +49,11 @@ def build_drivers_payload() -> List[Dict[str, Any]]:
             drivers_list.append(
                 {
                     "name": pkg_name,
-                    "source": "distro"
-                    if pkg_info.get("from_distro", False)
-                    else "third-party",
+                    "source": (
+                        "distro"
+                        if pkg_info.get("from_distro", False)
+                        else "third-party"
+                    ),
                     "free": bool(pkg_info.get("free", False)),
                     "builtin": bool(pkg_info.get("builtin", False)),
                 }
@@ -130,7 +132,7 @@ class DriversService(dbus.service.Object):
         return True
 
     @dbus.service.method(BUS_NAME, in_signature="", out_signature="aa{sv}")
-    def drivers(self) -> List[Dict[str, Any]]:
+    def drivers(self) -> dbus.Array:
         """Return the list of devices and their available drivers."""
 
         self._touch()
@@ -142,7 +144,7 @@ def main() -> None:
 
     dbus.mainloop.glib.DBusGMainLoop(set_as_default=True)
     bus = dbus.SessionBus()
-    dbus.service.BusName(DriversService.BUS_NAME, bus)
+    bus.request_name(DriversService.BUS_NAME)
 
     loop = GLib.MainLoop()
     DriversService(bus, mainloop=loop)
--- a/README.md
+++ b/README.md
@@ -54,6 +54,41 @@ The three main functions are:
 These functions only use python-apt. They do not need any other dependencies,
 root privileges, D-BUS calls, etc.
 
+## D-Bus API
+
+This project also provides a session-bus D-Bus service that exposes driver
+information. The service registers as `org.ubuntu.Drivers` on the session bus
+with the object path `/org/ubuntu/Drivers` and exposes a single method:
+
+* `drivers`: Returns a list of devices and their available drivers. The first
+  driver entry in each list is the recommended one.
+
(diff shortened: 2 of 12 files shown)


MitchellAugustin (changes requested):
Looks good for the most part, and worked in my quick tests[1] - my only specific requests for changes are around documentation. Please add docstrings that explain any functions whose purposes might not be immediately obvious (some of which I pointed out inline).

More broad question, what will the first user of this interface be? I am assuming this is primarily driven by planned improvements to the driver management interface GUI, but am not completely sure.

Also, regarding NVIDIA drivers specifically - it might be valuable to implement a dbus endpoint that would allow you to filter `gpgpu` drivers specifically, if you want to match up with how a user might use u-d-c on the command line[0] (ex: via `sudo ubuntu-drivers list --gpgpu` if they want to only list the server (gpgpu) driver variants, as opposed to `sudo ubuntu-drivers list` which is the default option that will list the desktop variants).
That one is totally optional though - my team does not have a need for this in a dbus interface today, so I'm just pointing it out in case it might be valuable for you if the dbus interface also exposes the exact same filtering options as the CLI.

[0] https://documentation.ubuntu.com/server/how-to/graphics/install-nvidia-drivers/
[1] tested on dgx b200:
```
ubuntu@buzzsaw:~/ubuntu-drivers-common$ gdbus introspect --session --dest org.ubuntu.Drivers --object-path /org/ubuntu/Drivers
node /org/ubuntu/Drivers {
  interface org.freedesktop.DBus.Introspectable {
    methods:
      Introspect(out s arg_0);
    signals:
    properties:
  };
  interface org.ubuntu.Drivers {
    methods:
      drivers(out aa{sv} arg_0);
    signals:
    properties:
  };
};
ubuntu@buzzsaw:~/ubuntu-drivers-common$ gdbus call --session --dest org.ubuntu.Drivers --object-path /org/ubuntu/Drivers --method org.ubuntu.Drivers.drivers 
([{'sys_path': <'/sys/devices/pci0000:15/0000:15:01.0/0000:16:00.0/0000:17:02.0/0000:19:00.0/0000:1a:00.0/0000:1b:00.0'>, 'modalias': <'pci:v000010DEd00002901sv000010DEsd00001999bc03sc02i00'>, 'vendor': <'NVIDIA Corporation'>, 'model': <'GB100 [B200]'>, 'drivers': <[{'name': <'nvidia-driver-580-open'>, 'source': <'distro'>, 'free': <false>, 'builtin': <false>}, {'name': <'nvidia-driver-580'>, 'source': <'distro'>, 'free': <false>, 'builtin': <false>}, {'name': <'nvidia-driver-580-server'>, 'source': <'distro'>, 'free': <false>, 'builtin': <false>}, {'name': <'nvidia-driver-580-server-open'>, 'source': <'distro'>, 'free': <false>, 'builtin': <false>}, {'name': <'nvidia-driver-590'>, 'source': <'distro'>, 'free': <false>, 'builtin': <false>}, {'name': <'nvidia-driver-590-open'>, 'source': <'distro'>, 'free': <false>, 'builtin': <false>}, {'name': <'nvidia-driver-590-server'>, 'source': <'distro'>, 'free': <false>, 'builtin': <false>}, {'name': <'nvidia-driver-590-server-open'>, 'source': <'distro'>, 'free': <false>, 'builtin': <false>}, {'name': <'xserver-xorg-video-nouveau'>, 'source': <'distro'>, 'free': <true>, 'builtin': <true>}]>}],)
```

I'll also ping @mwhudson here, in case he would like to comment.

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,151 @@
...
+    apt_pkg.init_config()
+    apt_pkg.init_system()
+    try:
+        cache = apt_pkg.Cache(None)
+    except Exception as ex:
+        print(ex)
+        return []
+
+    devices = UbuntuDrivers.detect.system_device_drivers(
+        apt_cache=cache, sys_path=sys_path, freeonly=False
```
3v1n0: Would `freeonly` as an API parameter provide us any benefit (like faster lookup) or we prefer to filter the values later?

MitchellAugustin:
@jibel I'm closing this one since we have now merged https://github.com/canonical/ubuntu-drivers-common/pull/163, which I believe fully supersedes it. Please let me know if that's incorrect and this should be reopened for any reason.

PR #163, opened 2026-06-16
Add dbus interface (w/ Gio.Application)

ashuntu:
Building on @jibel's PR (https://github.com/canonical/ubuntu-drivers-common/pull/156), this refactors his work significantly to use `Gio.Application` to manage the interface, and updates the tests to use more similar patterns to what already exists in `test_ubuntu_drivers.py`.

I think a lot of the concerns from https://github.com/canonical/ubuntu-drivers-common/pull/156 should be addressed, although I haven't looked into what is needed to convert this to a system service (https://github.com/canonical/ubuntu-drivers-common/pull/156#discussion_r2891119753) yet.

---

UDENG-10579


UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,317 @@
...
+    """
+
+    def __init__(
+        self, app: Gio.Application, invocation: Gio.DBusMethodInvocation
+    ) -> None:
+        self._app = app
+        self._invocation = invocation
+        self._task = Gio.Task.new(None, None, None, None)
+        self._task.set_return_on_cancel(False)
+        self._task.connect("notify::completed", self._on_done)
```
3v1n0: I would use the gio task callback rather than this.

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,317 @@
...
+        invocation: The pending D-Bus method invocation to reply to.
+    """
+
+    def __init__(
+        self, app: Gio.Application, invocation: Gio.DBusMethodInvocation
+    ) -> None:
+        self._app = app
+        self._invocation = invocation
+        self._task = Gio.Task.new(None, None, None, None)
+        self._task.set_return_on_cancel(False)
```
3v1n0: I'd set to true, but since we've no cancellable is still a noop

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,317 @@
...
+        self._task.connect("notify::completed", self._on_done)
+
+    def start(self) -> None:
+        """Begin the async payload build in a worker thread."""
+        self._app.hold()
+        self._task.run_in_thread(self._run)
```
3v1n0: In case of concurrent calls, maybe we want to keep the value cached and just queue any further request rather than replicate the same request?

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,317 @@
...
+                "vendor": info.get("vendor", ""),
+                "model": info.get("model", ""),
+                "drivers": drivers_list,
+            }
+        )
+
+    return payload
+
+
+def _to_variant(value: Any) -> GLib.Variant:
```
3v1n0: Wondering if it would be just better to build a gvariant from the start rather than convert native values, unless we've an usage for the payload in the native form

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,317 @@
...
+    def __init__(self, idle_timeout_seconds: int = 300) -> None:
+        super().__init__(
+            application_id=DriversService.BUS_NAME,
+            flags=Gio.ApplicationFlags.IS_SERVICE,
+        )
+        self.set_inactivity_timeout(idle_timeout_seconds * 1000)
+        self._service = DriversService(app=self)
+
+    def do_activate(self) -> None:
+        # D-Bus clients activate the service via method calls; no GUI activation.
```
3v1n0: Do we still need to override this?

debian/control
```diff
@@ -47,6 +46,9 @@ Depends: ${python3:Depends},
  python3-apt,
  python3-xkit,
  python3-click,
+ python3-gi,
+ gir1.2-glib-2.0,
+ dbus,
```
3v1n0: Is it needed?

I feel it's already brought by the rest.

setup.py
```diff
@@ -41,6 +41,9 @@
         ("/usr/share/doc/ubuntu-drivers-common", ["README"]),
         ("/usr/lib/nvidia/", glob.glob("nvidia-installer-hooks/*")),
         ("/usr/lib/ubiquity/target-config", glob.glob("ubiquity/target-config/*")),
+        ("/usr/lib/systemd/user/", ["systemd/ubuntu-drivers.service"]),
```
3v1n0: We still want this as user service?

I think it could be just a system service with any adm user being able to call it.

ashuntu: I think we do, but I'll end up having to revert a lot of the Gio.Application code unfortunately :(

ashuntu:
`black` formatting runs clean for me locally, not sure why that's failing :thinking:

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,378 @@
...
+        ):
+            driver_list.append(
+                GLib.Variant(
+                    "a{sv}",
+                    {
+                        "name": GLib.Variant("s", pkg_name),
+                        "source": GLib.Variant(
+                            "s",
+                            (
+                                "distro"
```
sergio-costas: This syntax, in several lines, is complex to read. I would put it as a normal `if` if that goes in several lines.

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,378 @@
...
+                "vendor": GLib.Variant("s", info.get("vendor", "")),
+                "model": GLib.Variant("s", info.get("model", "")),
+                "drivers": GLib.Variant("av", driver_list),
+            }
+        )
+
+    return GLib.Variant(f"({_DRIVERS_SIGNATURE})", (device_list,))
+
+
+class _IdleManager:
```
sergio-costas, line 127: Please, add a comment here explaining that it's used to close the daemon after inactivity.

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,378 @@
...
+        )
+
+        try:
+            self._loop.run()
+        finally:
+            if self._service is not None and self._connection is not None:
+                self._service.unexport(self._connection)
+
+    def on_bus_acquired(self, connection: Gio.DBusConnection, _name: str) -> None:
+        self._service = DriversService(
```
sergio-costas: Why are you passing the methods instead of just passing a reference to the `_idle_mgr` itself? I think it would be more clear in the receiving class knowing that it's a idle_manager and calling the methods, rather than just calling two methods passed "randomly".

sergio-costas (changes requested):
Just some little details, nothing serious.

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,378 @@
...
+        self._cached_result: Optional[GLib.Variant] = None
+        self._pending_invocations: List[Gio.DBusMethodInvocation] = []
+        self._task_running = False
+
+    def export(self, connection: Gio.DBusConnection) -> None:
+        """Register the D-Bus object on *connection*."""
+        if self._object_registration_id is not None:
+            return
+
+        if hasattr(connection, "register_object_with_closures2"):
```
3v1n0, line 246: Maybe we can just bump the glib dependency to 2.88 rather than doing this?

Unless we care about older than 26.04.

UbuntuDrivers/service/drivers_service.py
3v1n0: Bt in general do we need this for just a dbus call?

I feel we can just keep a queue of invocations once a task has been started and eventually return to them once the task is over, no?

Hoever keep it as a separated class if you think it's cleaner.

3v1n0 (commented):
Some nits, but overall it seems good to me.

setup.py
3v1n0: Yeah, true... Sorry about that.

UbuntuDrivers/service/drivers_service.py
ashuntu: I think 26.04 would be the minimum target, since older Ubuntu versions still have software properties to manage drivers and the purpose of this service is for the App Center

UbuntuDrivers/service/drivers_service.py
ashuntu: I added a comment, and I personally like the separation but I could see it being consolidated to the service if you'd prefer

UbuntuDrivers/service/drivers_service.py
ashuntu: It looks like `register_object_with_closures2` comes from >=2.84

debian/control
```diff
@@ -58,6 +59,9 @@ Description: Detect and install additional Ubuntu driver packages
  .
   - a Python API for detecting driver packages for a particular piece of
     hardware or the whole system.
+ .
+   - a D-Bus service (org.ubuntu.Drivers) for querying driver information
```
sergio-costas: I think that it should be `com.ubuntu.Drivers` in every place. I checked with d-spy and we are using both `com.canonical` and `com.ubuntu` everywhere.

sergio-costas (changes requested):
Just a detail.

Also, I see a lot of "cosmetic" changes, like empty lines removed. I think that they should belong to a different PR.

ashuntu:
> Also, I see a lot of "cosmetic" changes, like empty lines removed. I think that they should belong to a different PR.

I can do that, it was just that the `black` commands were failing because of formatting issues.

sergio-costas (changes requested):
You forgot to rename the `org.ubuntu.Drivers.conf` and `org.ubuntu.Drivers.service` to `com.ubuntu.Drivers...`.

UbuntuDrivers/service/drivers_service.py
```diff
@@ -193,11 +187,10 @@ class DriversService:
 
     def __init__(
         self,
-        hold: Callable[[], None],
-        release: Callable[[], None],
+        idle_manager: _IdleManager,
     ) -> None:
-        self._hold = hold
-        self._release = release
+        self._hold = idle_manager.hold
```
sergio-costas: I think that if the idle_manager is a fully-fledged object, it should be stored in the object, and its methods be called on it, not this way. I mean: `self._idle_manager = idle_manager`. And then, if you want to do a hold or a release, do `self._idle_manager.hold()` or `self._idle_manager.release()`.

sergio-costas (changes requested):

UbuntuDrivers/service/drivers_service.py
ashuntu: Hm, thought I addressed that already but it must've slipped through

debian/control
MitchellAugustin: +1 to com.ubuntu.*

debian/control
```diff
@@ -26,7 +26,6 @@ Build-Depends: debhelper-compat (= 13),
  udev,
  pciutils,
  libdrm-dev,
- python3-dbus,
```
MitchellAugustin, line 28: Why do we drop this?

UbuntuDrivers/service/drivers_service.py
```diff
@@ -0,0 +1,368 @@
...
+
+        interface com.ubuntu.Drivers
+            method drivers() -> aa{sv}
+
+    Args:
+        idle_manager: Instance of _IdleManager to manage inactivity timeouts.
+    """
+
+    BUS_NAME = "com.ubuntu.Drivers"
+    OBJ_PATH = "/org/ubuntu/Drivers"
```
MitchellAugustin: doesn't this also need changing to /com/ubuntu/drivers?

MitchellAugustin (commented):
Structurally, I don't see any more blocking issues, and it looks like my main feedback in #156 has been addressed here. This also seems to work on my laptop and DGX A100.

I added two inline questions, so once those are addressed, I'll give this a +1.

I also think Foundations should give this a quick look before we merge (especially if this is a candidate for Resolute) - @mwhudson , would you be able to take a look?

Regarding getting this in Resolute: I am not personally opposed, since this seems purely additive and unlikely to cause regressions with existing parts of u-d-c there, but it will be up to the SRU team to approve that in the end. In my opinion, it'd be ideal to see the full solution working in Stonking before we propose that SRU, though.

MitchellAugustin:
It looks like Michael is out until July 20th. Given the nature of this change, it'll be ideal if he/Foundations can also review before we merge this for Stonking. I think that will still probably come before August, once everyone is back from mid-year break.

With that said, I discussed with @ashuntu  in MM, and it sounds like a PPA build will at least be sufficient to unblock the next steps from his/Desktop/snapd team's side, so I'll get that out before EOW.

MitchellAugustin:
Building now in this PPA: https://launchpad.net/~mitchellaugustin/+archive/ubuntu/ubuntu-drivers-proposed

debian/control
ashuntu: https://github.com/canonical/ubuntu-drivers-common/pull/156 originally used `python3-dbus`, but this PR relies on `Gio` and `GLib` instead so it shouldn't be needed anymore

MitchellAugustin (approved):
Tested again on my [laptop](https://pastebin.canonical.com/p/GVdggrqSHC/) via the PPA build. LGTM from my perspective now, thanks @ashuntu .

I'll merge this once Foundations gives it a +1.

mwhudson (changes requested):
I asked claude opus to review this and it had quite a lot of comments: https://gist.github.com/mwhudson/6cdefab4f32d0170b5d773bfc1ba81ef  -- I'm not an expert on dbus and haven't verified every single one of this but most of them make some amount of sense. I'll also get claude to fix these but I haven't done that yet.

mwhudson:
Here are claude's fixes https://github.com/mwhudson/ubuntu-drivers-common/tree/fix/dbus-service-review

MitchellAugustin:
@ashuntu I think this all looks good now and am ready to merge it if you are done making changes.

PR #167, opened 2026-07-29
Remove the vestigial quirks system

mwhudson:
The quirks system was a Jockey-era mechanism for applying per-hardware X.org config snippets: driver package maintainer scripts would call quirks-handler -e/-d, which matched quirk files against DMI data and wrote snippets to /usr/share/X11/xorg.conf.d.

It has been non-functional for over a decade: nothing calls quirks-handler any more (current nvidia-graphics-drivers packaging has no reference to it), and the only two shipped quirk files declare Handler "nvidia-current|nvidia-current-updates" - package names that left the archive when Ubuntu moved to nvidia-driver-NNN naming - so even a caller would match nothing. The last functional change to the code was in 2012.

Remove the Quirks Python package, the quirk data files, the quirks-handler script, and everything used only by them: the quirkreader test and its xorg.conf fixture, the generated-settings.py machinery in tests/run, the python3-xkit dependency (build and runtime), and the Quirks references in debian/rules, the pre-commit hook, mypy.ini, and test_static.py.

The bbswitch quirk handling inside gpu-manager and its tests is a separate mechanism and is untouched.


MitchellAugustin:
@mwhudson are we certain that the quirks system isn't used outside of ubuntu-drivers and the nvidia driver packaging? 

I was under the impression that it was used in some capacity on some OEM platforms. (but am not certain)

mwhudson:
Well this is the hard thing to figure out. I guess we could ask the OEM team?

PR #168, opened 2026-07-29
Remove deprecated ubiquity integration

mwhudson:
Ubiquity is firmly deprecated. Drop the target-config hook (31ubuntu_driver_packages) and the setup.py entry that installed it to /usr/lib/ubiquity/target-config; the deb picked it up from the wholesale usr/ install, so no packaging changes are needed.

The --package-list CLI option is kept: ubiquity used it, but it is public CLI surface that other installer frontends can use and it is covered by the test suite. Also update ARCHITECTURE_REVIEW.md so the diagram and file-contract table no longer reference ubiquity.


MitchellAugustin:
LGTM, thanks @mwhudson . Should I go ahead and merge?

MitchellAugustin (approved):

mwhudson:
Yes please (I cannot, which slightly surprises me, I thought I had permissions for this)

PR #169, opened 2026-08-05
Add DisplayLink USB detection plugin with vendor/model plugin metadata support

mkaspryk-synaptics:
## Summary

Adds detection of DisplayLink USB graphics devices and recommends the
`displaylink-driver` package when appropriate. To support this cleanly, the
detection-plugin API is extended so plugins can optionally report
human-readable `vendor`/`model` metadata, matching what is already available
for modalias-based driver detection.

### DisplayLink detection plugin
- New `detect-plugins/displaylink.py` that scans USB sysfs
  (`/sys/bus/usb/devices`) for DisplayLink devices (USB vendor `17e9`).
- Devices already handled by the in-kernel `udl` DRM driver are **excluded**
  by inspecting USB interface descriptors (`bInterfaceClass=0xff`,
  `bInterfaceSubClass=0x00`, `bInterfaceProtocol=0x00`), mirroring the match
  logic in `drivers/gpu/drm/udl/udl_drv.c`. This avoids recommending the
  proprietary package for devices that already work with the open-source
  kernel driver.
- Falls back to a synthesized model name (`USB product <id>`) when the sysfs
  `product` attribute is unavailable.

### Plugin metadata support (`UbuntuDrivers/detect.py`)
- `detect_plugin_packages()` now accepts an optional `plugin_metadata`
  output dict and understands a new dictionary return format from plugins:
  a mandatory `packages` key plus optional `vendor` and `model` entries.
  The existing list/set return format remains fully supported.
- `system_driver_packages()` propagates plugin `vendor`/`model` into the
  resulting `PackageInfo`.
- Adds input validation for the dict format (type checks for `vendor`,
  `model`, and individual package names) and deduplicates/sorts plugin
  package results for deterministic output.

### Packaging fix (`setup.py`)
- Filters out directories when installing `detect-plugins/*`, so only files
  are shipped into `/usr/share/ubuntu-drivers-common/detect`.

### Tests & docs
- Adds tests covering the new dict return format (vendor/model collection,
  package dedup/sort, invalid vendor/model/package validation) and a
  `DisplayLinkPluginTest` exercising the plugin against a fake sysfs tree
  (device detection, `udl` exclusion, product-name fallback, non-matching
  vendors, and the `detect()` return contract).
- Documents the new dictionary plugin return format in the README.
- Adds a changelog entry (`1:0.10.10`).

## Backwards compatibility

The plugin return-type change is additive: existing plugins returning a
`list` or `set` continue to work unchanged, and the return type of
`detect_plugin_packages()` is preserved (`pluginname -> [package, ...]`).

## Manual tests

The ubuntu-drivers package was built on Ubuntu 24.04 and tested on Ubuntu 24.04 and 26.04.

sc2
sc1

For proper detection of DisplayLink devices, there is a need to download the Synaptics APT Repository  (https://www.synaptics.com/products/displaylink-graphics/downloads/ubuntu) and then install it 
`sudo apt install ./Downloads/synaptics-repository-keyring.deb` 
Without it, ubuntu-drivers will not see the needed proprietary driver.
These steps are required as long as the Synaptics APT Repository is not added to the Ubuntu archive (multiverse).

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -58,6 +58,13 @@ class PackageInfo(TypedDict, total=False):
     metapackage: str
 
 
+class PluginMetadata(TypedDict, total=False):
+    """Optional metadata returned by a driver detection plugin."""
+
+    vendor: str
+    model: str
+
+
 system_architecture = ""
 lookup_cache = {}
 custom_supported_gpus_json = "/etc/custom_supported_gpus.json"
@@ -669,7 +676,7 @@ def _get_db_name(syspath: str, alias: str) -> Tuple[Optional[str], Optional[str]
     vendor = None
     model = None
     for line in out.splitlines():
-        (k, v) = line.split("=", 1)
+        k, v = line.split("=", 1)
         if "_VENDOR" in k:
             vendor = v
         if "_MODEL" in k:
@@ -779,7 +786,7 @@ def system_driver_packages(
                 "runtimepm": _is_runtimepm_supported(apt_cache, p, alias),
                 "open_preferred": _is_open_prefered(apt_cache, p),
             }
-            (vendor, model) = _get_db_name(syspath, alias)
+            vendor, model = _get_db_name(syspath, alias)
             if vendor is not None:
                 packages[p.name]["vendor"] = vendor
             if model is not None:
@@ -798,16 +805,39 @@ def system_driver_packages(
         for p in nvidia_packages:
             packages[p]["recommended"] = p == recommended
 
+    # Keep the existing package result returned by detect_plugin_packages(),
+    # while collecting optional metadata separately. This preserves
+    # compatibility with callers which expect a mapping of plugin names to
+    # package-name lists.
+    plugin_metadata: Dict[str, PluginMetadata] = {}
+    plugin_packages = detect_plugin_packages(
+        apt_cache,
+        plugin_metadata=plugin_metadata,
+    )
+
     # add available packages which need custom detection code
-    for plugin, pkgs in detect_plugin_packages(apt_cache).items():
+    for plugin, pkgs in plugin_packages.items():
+        metadata = plugin_metadata.get(plugin, {})
+
         for p in pkgs:
             try:
                 apt_p = apt_cache[p]
-                packages[p] = {
+
+                package_info: PackageInfo = {
                     "free": _is_package_free(apt_cache, apt_p),
                     "from_distro": _is_package_from_distro(apt_cache, apt_p),
                     "plugin": plugin,
                 }
+
+                vendor = metadata.get("vendor")
+                if vendor:
+                    package_info["vendor"] = vendor
+
+                model = metadata.get("model")
+                if model:
+                    package_info["model"] = model
+
+                packages[p] = package_info
             except KeyError:
                 logging.debug("Package %s plugin not available. Skipping." % p)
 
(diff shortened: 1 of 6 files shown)


MitchellAugustin:
@mwhudson what is your opinion on this proposal, given that it sounds like it is really only adding functionality for users of the Synaptics apt repository? I am hesitant since we don't typically do this type of thing for non-Ubuntu-controlled sources, unless Synaptics is planning to land the drivers being detected here into the Ubuntu archives in the near future.

> For proper detection of DisplayLink devices, there is a need to download the Synaptics APT Repository (https://www.synaptics.com/products/displaylink-graphics/downloads/ubuntu) and then install it
> sudo apt install ./Downloads/synaptics-repository-keyring.deb
> Without it, ubuntu-drivers will not see the needed proprietary driver.

However, on the other hand, it should technically have no impact or visibility to users who did not already install the Synaptics sources (unless they read this source).

MitchellAugustin:
Hi @mkaspryk-synaptics , apologies for any confusion from my last comment, I was just informed that you are actually working with another team in Canonical on this. We will try to get this reviewed soon.

detect-plugins/displaylink.py
```diff
@@ -0,0 +1,114 @@
...
+def _read_sysfs_value(path):
+    """Read and return a stripped string from a sysfs attribute file."""
+    try:
+        with open(path, encoding="ascii") as attribute:
+            return attribute.read().strip()
+    except (OSError, UnicodeError):
+        return None
+
+
+def _matches_udl_interface(device_path):
```
MitchellAugustin, line 28: Structurally, this seems fine.
I don't have the relevant hardware to test this, so I will delegate to the submitter to test that it works as intended on that hardware.

detect-plugins/displaylink.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+        if (iface_class == _UDL_INTERFACE_CLASS
+                and iface_subclass == _UDL_INTERFACE_SUBCLASS
+                and iface_protocol == _UDL_INTERFACE_PROTOCOL):
+            return True
+
+    return False
+
+
+def _detect_displaylink_device(sysfs_root=USB_SYSFS_ROOT):
```
MitchellAugustin, line 58: Looks correct (will return vendor/model only if a device with displaylink vendor id, not covered by the in-tree driver, is found).

UbuntuDrivers/detect.py
```diff
@@ -1708,24 +1750,69 @@ def detect_plugin_packages(
...
 
+            if valid_packages:
+                # Sort and deduplicate the result to ensure deterministic
+                # output if a plugin encounters more than one matching
+                # device.
+                packages[fname] = sorted(set(valid_packages))
+
+                if plugin_metadata is not None and metadata:
+                    plugin_metadata[fname] = metadata
+
```
MitchellAugustin, line 1815: This function looks good. Seems like it should not change the functionality of existing users AFAICT (only diff would be the deduplication, which should be fine even if any package names are getting duplicated).

tests/test_ubuntu_drivers.py
```diff
@@ -8584,6 +8693,128 @@ def test_linux_detection_names_chroot8(self):
             chroot.remove()
 
 
+class DisplayLinkPluginTest(unittest.TestCase):
```
MitchellAugustin, line 8696: These all make sense to me

MitchellAugustin (approved):
Thanks @mkaspryk-synaptics . This looks technically correct to me, and it looks like you have taken care to keep things backwards-compatible so that nothing breaks with other device detection.

+1 from my end, contingent on the relevant packages landing in the archives (and on @mwhudson giving his +1 too). If you could link the NEW package launchpad bug in here for the actual drivers once it is submitted so that we can track along, that'd be helpful.

mkaspryk-synaptics:
> Thanks @mkaspryk-synaptics . This looks technically correct to me, and it looks like you have taken care to keep things backwards-compatible so that nothing breaks with other device detection.
> 
> +1 from my end, contingent on the relevant packages landing in the archives (and on @mwhudson giving his +1 too). If you could link the NEW package launchpad bug in here for the actual drivers once it is submitted so that we can track along, that'd be helpful.

Thanks @MitchellAugustin, for the review. I'm currently working on building source packages. Here is our PPA repository for tests: https://launchpad.net/~synaptics-displaylink/+archive/ubuntu/displaylink-driver.  Along the way, there will be the next release of our displaylink-driver - 6.4.0. With that version, we'll create a needs-packaging bug. I'll publish that bug here.

PR #172, opened 2026-08-18
Add modalias fetching using dmidecode

aglinserer:
Some arm devices are not very verbose with the information available on the CPU/processor side. This PR extends the existing functionality with parsing of `dmidecode --string processor-<..>` and extends the system_modalias this way. This also allows targeting a family of devices in oem-situations where we are not able to directly influence the product family name.

This allows to use for example the following modalias in a package to target a DGX Spark GB10 device:
```
alias processor:version:GB10 ub_drv_test_processor_version_gb10
```

A built version of this and some packages to test is available in https://launchpad.net/~aglinserer/+archive/ubuntu/ubuntu-drivers. The test packages are built from https://github.com/aglinserer/ub-drv-tests.

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -194,6 +194,57 @@ def system_modaliases(sys_path: Optional[str] = None) -> Dict[str, str]:
 
         aliases[modalias] = path
 
+    aliases.update(_dmidecode_processor_modaliases())
+    return aliases
+
+
+def _get_dmidecode_string(value: str) -> str:
+    try:
+        proc = subprocess.run(
+            ["dmidecode", "--string", value],
+            capture_output=True,
+            text=True,
+            timeout=10,
+        )
+    except (OSError, subprocess.SubprocessError) as e:
+        logging.debug("_get_dmidecode_string(): could not run dmidecode: %s", e)
+        return ""
+
+    if proc.returncode != 0:
+        logging.debug(
+            "_get_dmidecode_string(): dmidecode exited with %s: %s",
+            proc.returncode,
+            proc.stderr.strip(),
+        )
+        return ""
+    return proc.stdout.strip()
+
+
+def _normalize(value: str) -> str:
+    """Normalize program output for comparison.
+
+    Removes spaces and replaces commas with nothing. This is a best-effort
+    attempt to make the output of commands more comparable across different
+    systems. It is not guaranteed to be perfect, but it should be good enough
+    for most cases.
+    """
+    value = value.replace(" ", "").replace(",", "")
+    return value
+
+
+def _dmidecode_processor_modaliases() -> Dict[str, str]:
+    """Collect processor specific aliases using dmidecode.
+
+    Uses `dmidecode --string processor-xx` as the source.
+    A failure yields an empty dictionary or a partially filled dictionary.
+    """
+    aliases: Dict[str, str] = {}
+    for entry in ["family", "manufacturer", "version", "frequency"]:
+        ret = _get_dmidecode_string(f"processor-{entry}")
+        if not ret:
+            continue
+        ret = _normalize(ret)
+        aliases["processor:" + entry + ":" + ret] = "dmidecode"
     return aliases
 
 
--- a/tests/test_ubuntu_drivers.py
+++ b/tests/test_ubuntu_drivers.py
@@ -7119,9 +7119,7 @@ class DmidecodeModaliasTest(unittest.TestCase):
             UbuntuDrivers.detect._get_dmidecode_string("processor-family"), ""
         )
 
-    @patch(
-        "UbuntuDrivers.detect.subprocess.run", side_effect=OSError("no dmidecode")
-    )
+    @patch("UbuntuDrivers.detect.subprocess.run", side_effect=OSError("no dmidecode"))
     def test_get_dmidecode_string_missing_binary(self, mock_run):
         """_get_dmidecode_string() returns '' when dmidecode is absent (OSError)"""
 
--- a/tests/test_ubuntu_drivers.py
+++ b/tests/test_ubuntu_drivers.py
@@ -207,7 +207,8 @@ class DetectTest(unittest.TestCase):
             self.assertGreater(len(res), 3)
             self.assertTrue(":" in list(res)[0])
 
-    def test_system_modaliases_fake(self):
+    @patch("UbuntuDrivers.detect._dmidecode_processor_modaliases", return_value={})
+    def test_system_modaliases_fake(self, mock_proc):
         """system_modaliases() for fake sysfs"""
 
         res = UbuntuDrivers.detect.system_modaliases(self.umockdev.get_sys_dir())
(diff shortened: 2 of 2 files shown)


aglinserer:
One potential issue I saw: Running `ubuntu-drivers list` does not list these new packages, because `dmidecode` requires sudo privileges on some of the executions. Since executing ubuntu-drivers for installing is checking for sudo, it is not a real execution issue, but might be misleading.

What is the preferred resolution for this? Simply add the sudo check to `list` as well?

MitchellAugustin:
> What is the preferred resolution for this? Simply add the sudo check to list as well?

@aglinserer yeah, we probably need to do that here. The only other option would be to have it display a warning telling users that their list output may be incomplete without sudo, but I don't think that's a good option since some users might be parsing the list output and only expecting driver names.

IIRC subiquity is using `list` to determine its list - have you double-checked to confirm that it is always running that as root?

tests/test_ubuntu_drivers.py
```diff
@@ -7033,6 +7034,199 @@ def test_get_linux_chroot(self):
...
+    @patch(
+        "UbuntuDrivers.detect.subprocess.run",
+        side_effect=subprocess.TimeoutExpired(cmd="dmidecode", timeout=10),
+    )
+    def test_get_dmidecode_string_timeout(self, mock_run):
+        """_get_dmidecode_string() returns '' when dmidecode times out"""
+
+        self.assertEqual(
+            UbuntuDrivers.detect._get_dmidecode_string("processor-frequency"), ""
+        )
```
MitchellAugustin, line 7139: +0 on the tests above this comment. They all look technically correct and shouldn't add substantial runtime to the suite, but I don't really know that we get a ton of meaningful coverage from having tests for entirely trivial stuff like making sure str.replace works on some already statically defined examples.

(more of a general complaint I have with AI-generated tests).

Deferring review on tests below since I have a change request on the implementation that may impact them.

UbuntuDrivers/detect.py
```diff
@@ -194,6 +194,57 @@ def system_modaliases(sys_path: Optional[str] = None) -> Dict[str, str]:
 
         aliases[modalias] = path
 
+    aliases.update(_dmidecode_processor_modaliases())
```
MitchellAugustin, line 202: It seems a bit misleading to append our dmidecode information onto a field returned by a function named `system_modaliases`, given that this information is not actually coming from the modalias file(s). (and even to use the `XB-Modaliases` field in the debian packages for this in general, tbh).

What if, instead, we created a new function that looks for our own [user-defined d/control field](https://www.debian.org/doc/debian-policy/ch-controlfields.html#user-defined-fields) named `XB-udc-CPUs` (or similar), then just define that as its own field in the packages that we need to key off CPUs?
(or alternatively `XB-udc-dmi-identifiers`, if you prefer one d/control field that is usable for various DMI idents.)

UbuntuDrivers/detect.py
```diff
@@ -194,6 +194,57 @@ def system_modaliases(sys_path: Optional[str] = None) -> Dict[str, str]:
...
+    Uses `dmidecode --string processor-xx` as the source.
+    A failure yields an empty dictionary or a partially filled dictionary.
+    """
+    aliases: Dict[str, str] = {}
+    for entry in ["family", "manufacturer", "version", "frequency"]:
+        ret = _get_dmidecode_string(f"processor-{entry}")
+        if not ret:
+            continue
+        ret = _normalize(ret)
+        aliases["processor:" + entry + ":" + ret] = "dmidecode"
```
MitchellAugustin: This function LGTM (but +1 on adding the sudo check for dmidecode)

MitchellAugustin (changes requested):
Thanks @aglinserer ! I have a slightly different preference for where we store this data in `d/control` described inline - let me know if you think there is an issue with my counterproposal there.

UbuntuDrivers/detect.py
MitchellAugustin: @mwhudson Curious if you agree with my suggestion here.

For context: some upcoming platforms do not have precise enough information exposed by the system modalias files that u-d-c currently uses as a key (defined in `XB-Modaliases` in d/control), and the vendor's preferred thing to check instead is a field that is exposed by `dmidecode`.

I think we should simultaneously submit a PR to the upstream kernel to expose this same info via the CPU modaliases, but since that will take a long time to trickle down into our kernels, would also prefer for us to be able to use the info from dmidecode as a direct key as Andreas has started implementing here.

So, mainly just would like to know if you agree with me here that this would be better suited for its own functions in u-d-c, or if appending the data to the system_modaliases() function (to keep things in the same place) is better. (or if you'd prefer some other option).

mwhudson: I think I would prefer not to be dishonest about where the data is coming from yes, so a separate function and field in the control file makes sense to me. It's a bit tempting to say that matching against Modaliases is more future proof but I think it's very unlikely that this data will appear in the same format in an actual modalias file so it will probably be _less_ future proof to go that way. I guess this means a change to dh_modaliases as well?

mwhudson:
> > What is the preferred resolution for this? Simply add the sudo check to list as well?
> 
> @aglinserer yeah, we probably need to do that here. The only other option would be to have it display a warning telling users that their list output may be incomplete without sudo, but I don't think that's a good option since some users might be parsing the list output and only expecting driver names.

I think a warning on stderr in this case would be appropriate. Maybe only if there are binary packages processed that match on dmidecode data but perhaps that's too much work.

> IIRC subiquity is using `list` to determine its list - have you double-checked to confirm that it is always running that as root?

The subiquity server runs as root, so this should be OK.

UbuntuDrivers/detect.py
aglinserer: @MitchellAugustin @mwhudson , to clarify what the preferred course is, before I get into it. From my understanding there are 2 potential routes:

1) Rework the information gathering and remove it from `system_modaliases`. Extend the information instead in ~`system_driver_packages`~, `system_device_specific_metapackages`, ~`system_gpgpu_driver_packages`~ which are the consumers of `system_modalias`. Change the Modalias section to e.g. `dmidecode-processor`, which would result in a `.modalias`-File with the following contents: `alias dmidecode-processor:version:GB10 ub_drv_test_processor_version_gb10`. This is quite a mouthful, but pretty verbose and seems future proof. But my experience here is limited, so I'll leave this decision to you.

2) Rework information gathering similar to 1) and adapt the `dh_modaliases` as well to account for a whole new `XB-` field (e.g. `XB-Dmidecode`?) and extend the code in the relevant bits (these relevant bits, I need to first figure out).

While 2) is unarguably cleaner, it is also touching more things, ~especially the `dh_modaliases` which has been last modified 14 years ago (so it is proven it works)~.

Edit: After reading this again, it seems I'd need to add a whole additional `dh_dmidecode` or something like this for route 2.

antlassagne: I like route 2. It seems cleaner to me, I am not comfortable with hacking something that is not a modalias in a feature that is only about modaliases and even named after them. We keep things we're they are supposed to be and where people will expect them to be.

Now, going back to the problem that this PR resolves. We'd like to identify the CPU vendor / product name of ARM devices. Can't that be contained in the cpu modalias? Would that be a kernel PR instead of an ubuntu-drivers-common one? Maybe it's worth  asking the kernel team what they think

MitchellAugustin: @aglinserer I would prefer something more like option (2), where you have a new field / code path for `dh_dmidecode`.

> Can't that be contained in the cpu modalias? Would that be a kernel PR instead of an ubuntu-drivers-common one? Maybe it's worth asking the kernel team what they think

@antlassagne I definitely would prefer this in the long run (and have it on my TODO list), but I don't want us to be blocked on it - and I think landing something in the upstream kernel and then getting it backported is going to take longer than the u-d-c change. (and the u-d-c change will be helpful to have in case we find more helpful dmidecode keys in the future)

mwhudson:
FWIW, claude seems to think it's not that hard to synthesize what the cpu modalias _would_ be if that was enabled in the kernel: https://claude.ai/code/artifact/69f207bb-d9f1-4823-b50d-78e2ea397b4e (search for "the move")

MitchellAugustin:
@mwhudson I do intend to work on a kernel patch for that too, I'm just concerned about the timeline on it actually landing in our kernels, hence the interest in having this in u-d-c sooner.
(and I also think a kernel patch that changes the fields exposed by the kernel to userspace might be unacceptable for an SRU, since many more userspace apps might depend on the current format - whereas with the proposed u-d-c changes, there would be no regression risk since it would depend on an entirely novel debian metadata field to pull any new content. So that part is probably better for a future Ubuntu release anyway.)

aglinserer:
Hi @MitchellAugustin @mwhudson , reworked this to be as separate as it can be IMO. I also agree with Antoine that we should merge #173 into this, albeit I am not fully certain why this MIDR detection requires a different behavior alltogether (have to check the PR closer)

MitchellAugustin:
FWIW: Followed up with Andreas, Michael, and Antoine in MM on why we need both MIDR and dmi for the different platforms. (it isn't necessarily obvious from the PRs alone)

MitchellAugustin (approved):
Thank you for the refactor @aglinserer !

This looks great now. You can also just drop the d/changelog update entirely to resolve the merge conflict (I'll need to write a new one with everyone's staged changes anyway).

UbuntuDrivers/detect.py
```diff
@@ -812,6 +874,20 @@ def system_driver_packages(
             except KeyError:
                 logging.debug("Package %s plugin not available. Skipping." % p)
 
+    dmidecode = dmidecode_aliases()
+    alias_map = apt_cache_map(apt_cache, key="Dmidecode")
+    for alias, _ in dmidecode.items():
+        for p in packages_for_modalias(apt_cache, alias, modalias_map=alias_map):
+            packages[p.name] = {
```
mwhudson, line 885: This dict lacks 'runtimepm' which I realize doesn't really make a lot of sense in this context but probably the shape of the data should be the same?

UbuntuDrivers/detect.py
```diff
@@ -812,6 +874,20 @@ def system_driver_packages(
...
+                "modalias": alias,
+                "syspath": "",
+                "free": _is_package_free(apt_cache, p),
+                "from_distro": _is_package_from_distro(apt_cache, p),
+                "recommended": True,
```
mwhudson, line 890: Does it make sense for all matching packages to have recommended: true? Again probably unlikely to matter in practice.

UbuntuDrivers/detect.py
```diff
@@ -812,6 +874,20 @@ def system_driver_packages(
             except KeyError:
                 logging.debug("Package %s plugin not available. Skipping." % p)
 
+    dmidecode = dmidecode_aliases()
```
mwhudson, line 877: So this branch adds this call to system_driver_packages. Should it be added to the other functions that call sytem_modaliases as well (system_device_specific_metapackages, system_gpgpu_driver_packages)?

debhelper/test_dh_dmidecode
```diff
@@ -0,0 +1,76 @@
...
+if echo "$header" | grep -qE '^[[:space:]]*Dmidecode:[[:space:]]+[^[:space:]]'; then
+    fail "Unexpected DMI aliases without a .dmialiases file in\n$header"
+fi
+
+# now provide a manual DMI alias file; aliases are grouped by package,
+# packages are sorted, and any extra tokens after the package are ignored
+cat <<EOF > debian/foo.dmialiases
+# machine specific metapackages
+alias dmidecode:processor:Version:AMDRyzenAI7PRO350w/Radeon860M oem-amd-meta
+alias dmidecode:processor:Version:GB10 oem-nvidia-meta ignorethisextrafield
```
mwhudson: So the `oem-nvidia-meta` here doesn't really mean anything right? In Modaliases it's a module name but here it looks like a package name but that doesn't really make sense because it's the package that this alias is added to that is being matched to. I guess we should drop this?

UbuntuDrivers/detect.py
mwhudson: Semi-relatedly there are some docstrings that I think need updating.

debhelper/test_dh_dmidecode
```diff
@@ -0,0 +1,76 @@
...
+echo "$header" | grep -q '^[[:space:]]*Package: foo$' || fail "No proper Package: header in\n$header"
+if echo "$header" | grep -qE '^[[:space:]]*Dmidecode:[[:space:]]+[^[:space:]]'; then
+    fail "Unexpected DMI aliases without a .dmialiases file in\n$header"
+fi
+
+# now provide a manual DMI alias file; aliases are grouped by package,
+# packages are sorted, and any extra tokens after the package are ignored
+cat <<EOF > debian/foo.dmialiases
+# machine specific metapackages
+alias dmidecode:processor:Version:AMDRyzenAI7PRO350w/Radeon860M oem-amd-meta
```
mwhudson: I think this (and the other bits) should "version", not "Version" ?

mwhudson (commented):
Here are some nits and tweaks I think I'd like to see addressed before merge. Basically fine though!

UbuntuDrivers/detect.py
```diff
@@ -812,6 +874,20 @@ def system_driver_packages(
...
+    alias_map = apt_cache_map(apt_cache, key="Dmidecode")
+    for alias, _ in dmidecode.items():
+        for p in packages_for_modalias(apt_cache, alias, modalias_map=alias_map):
```
mwhudson, line 880: This loop should probably do the freeonly / metapackage filtering the loop over `packages_for_modalias` does?

aglinserer: Added the extra checks in 2687f08

UbuntuDrivers/detect.py
aglinserer: Aligned the data dict in 69f0b9b

UbuntuDrivers/detect.py
aglinserer: This I copied from above, but I think it makes sense. Since these packages are targeted for the system and intended to be system optimizations, not having them does not really benefit them.

UbuntuDrivers/detect.py
aglinserer: system_device_specific_metapackages - only called in list-oem, which feels more like dead code to me within this repo. Is list-oem called by subiquity or something similar? Then it can make sense, but at this point I am more confused by the actual distinction between list and list-oem.
system_gpgpu_driver_packages - it is not really related to gpgpu IMO, but if you wish I can add it there to.

debhelper/test_dh_dmidecode
aglinserer: You are right. Fixed in 3cc7f9f

debhelper/test_dh_dmidecode
aglinserer: In the modalias usage we have done so far, this has always been the name of the package to install. I am not 100% certain how this is resolved and need to check it again. I think it is safe to be dropped, but I need to test this and also adjust the perl parser, which right now is almost a copy of the dh_modalias version.

debhelper/test_dh_dmidecode
```diff
@@ -0,0 +1,76 @@
...
+alias dmidecode:processor:version:GB10 oem-nvidia-meta ignorethisextrafield
+alias dmidecode:processor:version:GB300 oem-nvidia-meta
+EOF
+
+# build it and check the generated Dmidecode header
+dpkg-buildpackage -us -uc
+
+header=`dpkg -I ../foo_1_*.deb`
+echo "$header" | grep -q '^[[:space:]]*Package: foo$' || fail "No proper Package: header in\n$header"
+echo "$header" | grep -q '^[[:space:]]*Dmidecode: oem-amd-meta(dmidecode:processor:version:AMDRyzenAI7PRO350w/Radeon860M), oem-nvidia-meta(dmidecode:processor:version:GB10, dmidecode:processor:version:GB300)$' || fail "No proper DMI aliases in\n$header"v
```
mwhudson, line 74: I don't think you want the "v" at the end here :-)

UbuntuDrivers/detect.py
mwhudson: I think list-oem is called by subiquity, yes. I agree gpgpu is probably not going to need this but it would be super confusing if it was needed and simply not called (I hate how much almost duplicate code there is between these endpoints!)

debhelper/test_dh_dmidecode
mwhudson: I guess this would be annoying to implement as the parsing code in the python code would have to change but I'm also hesitant to meaningless stuff in the control files. @MitchellAugustin what do you think?

MitchellAugustin: In my opinion: if it is known that this field would always be useless, and if cleaning it up wouldn't make this harder to SRU for any reason, I agree we should try to clean it up. Otherwise it is a bit more debatable.

MitchellAugustin:
@aglinserer I think you might need to rebase on latest master, or if you have done that already, correct the misspelled `_is_open_prefered` that was recently fixed there. (which is causing the CI failure)

debhelper/test_dh_dmidecode
aglinserer: I made the change to the parser, so the module name is no longer required. I did change it in a way that it is now deriving the packagename from the filename and putting it there instead of the module name. This has the benefit of not requiring adjustments in the module parser logic within detect.py:apt_cache_map.

Let me know if this approach is acceptable.

debhelper/dh_dmidecode
```diff
@@ -0,0 +1,108 @@
...
+=head1 PACKAGE DMI ALIAS FILES
+
+The DMI aliases must be provided manually in a package file named
+B<debian/>I<package>B<.dmialiases>.
+
+Each non-comment line must contain an alias and the package providing it:
+
+=over 4
+
+alias <dmi-alias> <package>
```
mwhudson: So this documentation is stale now.

UbuntuDrivers/detect.py
```diff
@@ -198,6 +198,62 @@ def system_modaliases(sys_path: Optional[str] = None) -> Dict[str, str]:
...
+        logging.debug(
+            "_get_dmidecode_string(): dmidecode exited with %s: %s",
+            proc.returncode,
+            proc.stderr.strip(),
+        )
+        return ""
+    return proc.stdout.strip()
+
+
+def _normalize(value: str) -> str:
```
mwhudson: I feel like this function is perhaps a bit over-broadly named. Maybe `_normalize_dmidecode_value` or something? I'm also a little confused as to how these escaping rules interact with the parsing `apt_cache_map` does, I think trying to match on a value containing parentheses would go badly, for example?

mwhudson (approved):
This is very close now, thanks (and sorry for all the round-trips). Just a couple of quibbles, I don't think I need to look at this again (happy to merge it when ready if needed).

UbuntuDrivers/detect.py
aglinserer: Renamed and adjusted to now remove all characters used in `apt_cache_map`. You are right, this would've led to issues.

debhelper/dh_dmidecode
aglinserer: Removed the obsolete `<package>` mention

MitchellAugustin:
@aglinserer looks like this now needs a rebase and one linter issue fix

MitchellAugustin:
Thanks @aglinserer ! Tests passed for me with your PPA on Spark and my Intel CPU laptop (thanks for preparing those!)

Merging now.

PR #173, opened 2026-08-25
Add MIDR bit detection and tests (LP: #2165071)

MitchellAugustin:
Add the ability to install packages automatically based on a matching MIDR register.

For some upcoming platforms, the existing identification options in ubuntu-drivers for installing platform-specific metapackages are not specific enough.

This PR allows for `hwe` and `oem` metapackages to declare an `XB-Udc-Midr` field (which debhelper strips to `Udc-Midr` in the binary) containing the MIDR value for a CPU that they wish for it to be installed on by default.

This was tested by creating a local repository with an apt package whose Udc-Midr value matches the one on my system:
```
$ apt show hwe-test-platform-server-meta
Package: hwe-test-platform-server-meta
Version: 0.1ubuntu1
Priority: optional
Section: admin
Source: test-server-defaults
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Installed-Size: 8,192 B
Depends: test-server-common
Udc-Midr: 0x00000000410fd4f0
Download-Size: 1,702 B
APT-Sources: file:/srv/test-server-defaults-archive resolute/main arm64 Packages
Description: test-platform metapacakge
```
On this host:
```
$ cat /sys/devices/system/cpu/cpu0/regs/identification/midr_el1
0x00000000410fd4f0
```

I confirmed that `sudo ubuntu-drivers list`, `sudo ubuntu-drivers list-oem`, and `sudo ubuntu-drivers install` all pulled the new metapackage as desired. I also confirmed that, when I changed my midr_el1 value to something that didn't match (via bind mount), ubuntu-drivers correctly did not queue the metapackage for installation.


antlassagne:
This should most probably be coordinated with https://github.com/canonical/ubuntu-drivers-common/pull/172 since it seems highly related

UbuntuDrivers/detect.py
```diff
@@ -64,6 +66,33 @@ class PackageInfo(TypedDict, total=False):
...
+    variant: int
+    architecture: int
+    part_number: int
+    revision: int
+
+
+def parse_midr(value: str) -> Optional[MidrInfo]:
+    """Parse a MIDR string into the fields defined by MIDR_EL1."""
+    try:
+        midr = int(value, 0)
```
mwhudson, line 80: I think this is over broad, it accepts things like "10", "0o777", "0b11101" etc. I think we should probably require a 0x prefix and parse with int(,16).

UbuntuDrivers/detect.py
```diff
@@ -331,6 +392,43 @@ def apt_cache_modalias_map(
...
+    midr_map: Dict[MidrInfo, Set[str]] = {}
+
+    for package in apt_cache.packages:
+        try:
+            candidate = depcache.get_candidate_ver(package)
+            records.lookup(candidate.file_list[0])
+            values = records["Udc-Midr"]  # Defined as XB-Udc-Midr
+            if not values:
+                continue
+        except (KeyError, AttributeError, UnicodeDecodeError):
```
mwhudson, line 410: How do you expect to get AttributeError here? It's usually the result of a bug. Also I'd prefer records.get(..., "") vs catching the KeyError (exceptions should, generally speaking, be exceptional)

UbuntuDrivers/detect.py
```diff
@@ -197,6 +226,38 @@ def system_modaliases(sys_path: Optional[str] = None) -> Dict[str, str]:
     return aliases
 
 
+def system_midr(sys_path: Optional[str] = None) -> Dict[str, str]:
```
mwhudson, line 229: Super pedantic but this should probably be system_midrs (i.e. the plural) right?

UbuntuDrivers/detect.py
```diff
@@ -408,6 +506,22 @@ def packages_for_modalias(
     return [apt_cache[p] for p in pkgs]
 
 
+def packages_for_midr(
+    apt_cache: apt_pkg.Cache,
+    midr: str,
+    midr_map: Optional[Dict[MidrInfo, Set[str]]] = None,
```
mwhudson, line 512: All callers pass this argument I think. Is there any benefit to being conditional about it?

UbuntuDrivers/detect.py
```diff
@@ -964,6 +1097,24 @@ def system_device_specific_metapackages(
...
+                ):
+                    continue
+                packages[p.name] = {
+                    "midr": midr,
+                    "syspath": syspath,
+                    "free": _is_package_free(apt_cache, p),
+                    "from_distro": _is_package_from_distro(apt_cache, p),
+                    "recommended": True,
+                    "support": _pkg_get_support(apt_cache, p),
+                    "open_preferred": _is_open_prefered(apt_cache, p),
```
mwhudson, line 1115: As in the other branch, does recommended: true on every package found this way make sense? Should it add runtimepm: false for consistency?

UbuntuDrivers/detect.py
```diff
@@ -964,6 +1097,24 @@ def system_device_specific_metapackages(
                 "open_preferred": _is_open_prefered(apt_cache, p),
             }
 
+    if midrs:
+        midr_map = apt_cache_midr_map(apt_cache)
+        for midr, syspath in midrs.items():
+            for p in packages_for_midr(apt_cache, midr, midr_map=midr_map):
+                if not fnmatch.fnmatch(p.name, "oem-*-meta") and not fnmatch.fnmatch(
```
mwhudson, line 1104: I can't help but notice that this does _not_ exclude hwe-*-meta. But that's not a fix for this branch.

mwhudson (commented):
Here's some fairly small comments.

One other thing is that as nice as `MidrInfo` is, nothing in this branch ever reads an attribute from it. I think you could pass it around as an `int` everywhere with no loss of functionality. If we decide to keep it, can we use it everywhere and not bounce between MidrInfos and strings so much?

Another is that I'm not really sold on Udc-Midr: as a field name. In particular I don't think it makes sense to have the Udc bit -- too much mechanism not enough policy. `Matching-Midrs:` ? Not sure.

UbuntuDrivers/detect.py
mwhudson: Ah I see this is copying house style. I hate it but if we fix it then we should fix it everywhere.

MitchellAugustin:
> as nice as MidrInfo is, nothing in this branch ever reads an attribute from it. I think you could pass it around as an int everywhere with no loss of functionality. If we decide to keep it, can we use it everywhere and not bounce between MidrInfos and strings so much?

@mwhudson good point. When I started on this, I was originally going to allow users to match on individual components of the MIDR value, and then my implementation ended up drifting a bit and I only ended up matching exact strings. However, I think if I'm going to change the fields anyway, I could add that back in. What do you think about:

`Matching-Midrs`: Exact matches for the full MIDR string
`Matching-Midr-Implementers`: Exact matches for the MIDR implementer
`Matching-Midr-Variant`: Exact matches for MIDR variant
`Matching-Midr-Architecture`: Exact matches for MIDR architecture
`Matching-Midr-PartNum`: Exact matches for MIDR part number
`Matching-Midr_Revision`: Exact matches for MIDR revision

Then, if multiple of these are specified, we would only match on platforms that match all of them.

Our use case does not need this, so I am alternatively fine with just simplifying it, dropping MidrInfo, and matching on the raw int if you feel the above is overengineered - but I could forsee the above solution being useful someday.

UbuntuDrivers/detect.py
MitchellAugustin: Since I want this PR to be as SRUable as possible, I'll change it here but do the changes to instances of this pattern elsewhere in a future PR.

MitchellAugustin: @mwhudson actually, on a deeper look, this is a bit weirder. records is a `PackageRecords` object, which does not define `get()`. The [docs](https://apt-team.pages.debian.net/python-apt/library/apt_pkg.html) for apt-pkg define TagSection() for this, but also say that it is deprecated since we can do the lookup on the object itself , which looks to be what the house style is doing here. With that in mind, do you think this should still be changed?

UbuntuDrivers/detect.py
MitchellAugustin: Yes. We're following the model set in `system_device_specific_metapackages`:
```
'recommended': Always True; we always recommend you install these
                     packages.
```

since we do recommend them for users of platforms they match.

runtimepm is not a relevant attribute for these hwe metapackages, so we should not define it here. (same as in `system_device_specific_metapackages`, where it is also undefined.)

UbuntuDrivers/detect.py
MitchellAugustin: I just added the null check and default assignment on the offchance that a caller implemented in the future does not pass a midr_map, but I guess it is a bit overprotective since reviews on such a hypothetical PR would probably catch that.

I don't think there's any harm in how it is implemented but can remove it and/or force the caller to pass a midr_map if you prefer.

UbuntuDrivers/detect.py
MitchellAugustin: @mwhudson not sure I understand, the full line is
```
                if not fnmatch.fnmatch(p.name, "oem-*-meta") and not fnmatch.fnmatch(
                    p.name, "hwe-*-meta"
                ):
```
it just ended up structured weirdly because of the linter

mwhudson:
> Our use case does not need this, so I am alternatively fine with just simplifying it, dropping MidrInfo, and matching on the raw int if you feel the above is overengineered - but I could forsee the above solution being useful someday.

I think this is overengineered unless you can give me a realistic use case (you're certainly not going to want to match on variant or partnum alone are you?)

MitchellAugustin:
@mwhudson actually, now that I'm looking at this again, our use case actually originally only called for matching on implementer + partnum, so matching exactly on all other fields might actually be overly tight for our use case. (it works on the machines we have access to, but originally the partner just pointed us at implementer+partnum).

MitchellAugustin:
@mwhudson I made some changes based on your suggestion in our messages:

`Midr: implementer:0x4e,part:0x010` (or any ordering or selection of the MIDR fields) is now the format. I think this is preferable to allowing raw full MIDR strings to be set in d/control, since it requires the packager to think about the meaning of the specific fields in the MIDR value that they are trying to match on, and enumerate those explicitly.

This change introduces a slightly different MidrFields element (which represents this partial specification). I renamed `MidrInfo` to `SystemMidr` to disambiguate - `SystemMidr` is for a full system MIDR read from the runner's machine, and I added `MidrFields` as a separate one to emphasize that it can intentionally have some fields undefined.

Let me know if you have any questions on this. Thanks!

UbuntuDrivers/detect.py
mwhudson: Argh well let's leave this then. but also let's come back and try to reduce some of the duplication in this codebase.

UbuntuDrivers/detect.py
```diff
@@ -198,6 +286,38 @@ def system_modaliases(sys_path: Optional[str] = None) -> Dict[str, str]:
...
+    for path, dirs, files in os.walk(cpus):
+        if "midr_el1" not in files:
+            continue
+
+        midr_path = os.path.join(path, "midr_el1")
+        try:
+            with open(midr_path) as midr_file:
+                midr = midr_file.read().strip()
+        except IOError as e:
+            logging.debug("system_midrs(): Cannot read %s: %s", midr_path, e)
```
mwhudson, line 305: I wonder about logging at debug for this. It feels higher priority than that in some ways but OTOH spamming it at the user every time ubuntu-drivers runs is not showing them anything they have any power to do anything about.

UbuntuDrivers/detect.py
```diff
@@ -198,6 +286,38 @@ def system_modaliases(sys_path: Optional[str] = None) -> Dict[str, str]:
...
 
 
+def system_midrs(sys_path: Optional[str] = None) -> Dict[str, str]:
+    """Get unique MIDR values present in the system.
+
+    Return a MIDR value → sysfs path map.
+    """
+    midrs = {}
+    cpus = f"{sys_path}/devices/system/cpu" if sys_path else "/sys/devices/system/cpu"
+    for path, dirs, files in os.walk(cpus):
```
mwhudson, line 296: Claude points out that a glob on /sys/devices/system/cpu/*/midr_el1 (or whatever) is more efficient than this.

UbuntuDrivers/detect.py
```diff
@@ -787,6 +963,23 @@ def system_driver_packages(
...
 
+    if midrs:
+        midr_map = apt_cache_midrs_map(apt_cache)
+        for midr, syspath in midrs.items():
+            for p in packages_for_midr(apt_cache, midr, midr_map=midr_map):
+                if freeonly and not _is_package_free(apt_cache, p):
+                    continue
+                if not include_oem and fnmatch.fnmatch(p.name, "oem-*-meta"):
+                    continue
+                packages[p.name] = {
```
mwhudson, line 974: What should this do if the package has already matched via another mechanism? Very edge-casey but probably worth thinking about?

mwhudson (commented):
OK I think I like this version a lot more, thanks for the changes. Only one thing this absolutely needs before merge: documentation!

UbuntuDrivers/detect.py
```diff
@@ -370,7 +524,7 @@ def package_get_nv_allowing_driver(did: str) -> Optional[str]:
 def packages_for_modalias(
     apt_cache: apt_pkg.Cache,
     modalias: str,
-    modalias_map: Optional[Dict[str, Tuple[Any, Dict[str, Set[str]]]]] = None,
+    modalias_map: Optional[Dict[Any, Tuple[Any, Dict[str, Set[str]]]]] = None,
```
aglinserer: I dislike moving from `str` to `Any`. Why not list the things that can be a key?

UbuntuDrivers/detect.py
```diff
@@ -65,6 +77,82 @@ class PackageInfo(TypedDict, total=False):
...
+
+    if not 0 <= midr <= 0xFFFFFFFF:
+        return None
+
+    return SystemMidr(
+        implementer=(midr >> 24) & 0xFF,
+        variant=(midr >> 20) & 0xF,
+        architecture=(midr >> 16) & 0xF,
+        part_number=(midr >> 4) & 0xFFF,
+        revision=midr & 0xF,
```
aglinserer: Super nitpicky, but you already list the max values (MIDR_FIELD_MAX) so you could use those and reduce the amount of magic numbers.

UbuntuDrivers/detect.py
aglinserer: Would it be viable to update the entry if already existing?

UbuntuDrivers/detect.py
```diff
@@ -965,6 +1159,24 @@ def system_device_specific_metapackages(
...
 
+    if midrs:
+        midr_map = apt_cache_midrs_map(apt_cache)
+        for midr, syspath in midrs.items():
+            for p in packages_for_midr(apt_cache, midr, midr_map=midr_map):
+                if not fnmatch.fnmatch(p.name, "oem-*-meta") and not fnmatch.fnmatch(
+                    p.name, "hwe-*-meta"
+                ):
+                    continue
+                packages[p.name] = {
```
aglinserer: Same as above applies - if the package already exists it should be updated instead (should not make a difference in the greater picture, but is cleaner IMO)

UbuntuDrivers/detect.py
```diff
@@ -409,6 +563,27 @@ def packages_for_modalias(
...
+    if midr_map is None:
+        midr_map = apt_cache_midrs_map(apt_cache)
+
+    parsed_midr = parse_midr(midr)
+    if parsed_midr is None:
+        return []
+
+    packages = set()
+    for fields, matching_packages in midr_map.items():
+        if all(getattr(parsed_midr, name) == value for name, value in fields):
```
aglinserer: This line is heavy IMO. Shouldn't `all(parsed_midr[name] == value for name, value in fields)` suffice and save you from the `getattr` call?

UbuntuDrivers/detect.py
MitchellAugustin: It should set the same fields either way, as it does currently (with the exception of the "midr" or "modalias" field that indicates what detection criterion it was associated with). If detected by both, the modalias field would be shadowed since the packages entry would be overwritten with the one containing `midrs`, but given that it is deterministically ordered due to `system_midrs` coming last, I don't think that's an issue. Let me know if you disagree.

aglinserer (commented):
I added some comments and minor nitpicks that stood out to me

UbuntuDrivers/detect.py
MitchellAugustin: I think it probably is not worth doing at a higher log level. This seems like an already quite unlikely scenario to happen (since it will most primarily be used via the Ubuntu installer in a privileged context, so permissions should rarely be an issue (and on my VR machines, midr_el1 is globally readable anyway)), and even if it _does_  occur, it won't necessarily cause a real problem for a user (ex: if they're on an arm machine that has no matching packages via MIDR anyway) - so for users like that, it'd be both a non-actionable and also non-meaningful warning.

I'd also assume that any packages which _do_ match on a MIDR value would also be maintained by someone who has tested to ensure that the MIDR value is correctly readable by u-d-c on at least one version of that platform. In that case, being able to ask for or collect debug logs in the even of an issue should suffice IMO.

UbuntuDrivers/detect.py
MitchellAugustin: agreed, fixed

UbuntuDrivers/detect.py
MitchellAugustin: No, since parsed_midr is a SystemMidr and not a dict. We'd need to convert it with `parsed_midr._asdict()` here to reference it like that.

UbuntuDrivers/detect.py
MitchellAugustin: Fair point, will look at that

UbuntuDrivers/detect.py
MitchellAugustin: I think this might have been a mistake. Will fix.

UbuntuDrivers/detect.py
MitchellAugustin: fixed

UbuntuDrivers/detect.py
MitchellAugustin: I ended up fixing this

MitchellAugustin:
@mwhudson @aglinserer lmk what you think, all comments should be addressed now

UbuntuDrivers/detect.py
mwhudson: As sorry i meant all the log messages in this function, including e.g. the malformed headers. but yes, let's leave it as is for now.

mwhudson (approved):

MitchellAugustin:
Thanks for the reviews @mwhudson @aglinserer ! Ran one last test with the new format and confirmed it works on a real machine. Merging!

PR #175, opened 2026-08-27
feat: Add extra_packages field to dbus output

ashuntu:
Adds a new field to the dbus service output called `packages` and `gpgpu_packages` that contains the other packages `ubuntu-drivers-common` already checks for and installs alongside drivers. This should let consumers know which other packages to install alongside NVIDIA drivers specifically, since they often can come with dkms packages, etc.

Related to https://github.com/canonical/ubuntu-drivers-common/pull/174

---

UDENG-11309

### DIFF
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -32,6 +32,7 @@ class DriverInfo(TypedDict, total=False):
     recommended: bool
     support: Optional[str]
     open_preferred: bool
+    metapackage: str
 
 
 class DeviceInfo(TypedDict, total=False):
@@ -1114,6 +1115,8 @@ def system_device_drivers(
       'open_preferred': Boolean flag whether the "open" kernel module variant
                      is preferred over the closed one for this package, per
                      _is_open_preferred().
+      'metapackage': Name of the associated metapackage (e.g.
+                     "nvidia-headless-no-dkms-470"), if any.
     """
     result: Dict[str, DeviceInfo] = {}
     if not apt_cache:
@@ -1143,6 +1146,8 @@ def system_device_drivers(
             drivers[pkg]["support"] = pkginfo["support"]
         if "open_preferred" in pkginfo:
             drivers[pkg]["open_preferred"] = pkginfo["open_preferred"]
+        if "metapackage" in pkginfo:
+            drivers[pkg]["metapackage"] = pkginfo["metapackage"]
 
     # now determine the manual_install device flag: this is true iff all driver
     # packages are "manually installed"
--- a/UbuntuDrivers/detect.py
+++ b/UbuntuDrivers/detect.py
@@ -1310,6 +1310,7 @@ def _build_installation_list(
     include_dkms: bool,
     gpgpu: bool = False,
     filter_installed: bool = True,
+    skip_runtimepm_marker: bool = False,
 ) -> List[str]:
     """
     Build the list of packages to install including metapackages and modules.
@@ -1326,6 +1327,8 @@ def _build_installation_list(
             modules/lrm-meta "already installed" check below no longer
             prevents driver_found from being set, so the loop still stops at
             the first resolved driver (aside from hwe- metas).
+        skip_runtimepm_marker: Boolean, if True, skip creating
+            /run/nvidia_runtimepm_supported
 
     Returns:
         List of package names to install including metapackages and module packages.
@@ -1342,7 +1345,7 @@ def _build_installation_list(
         if not p.startswith("hwe-") and driver_found:
             continue
 
-        if not gpgpu:
+        if not gpgpu and not skip_runtimepm_marker:
             candidate_ver = depcache.get_candidate_ver(cache[p])
             records = apt_pkg.PackageRecords(cache)
             records.lookup(candidate_ver.file_list[0])
@@ -1436,6 +1439,7 @@ def already_installed_filter(
     include_dkms: bool,
     gpgpu: bool = False,
     filter_installed: bool = True,
+    skip_runtimepm_marker: bool = False,
 ) -> List[str]:
     """
     Sort driver branch to install according to preference, then select
@@ -1451,6 +1455,8 @@ def already_installed_filter(
         filter_installed: Boolean, if False, the returned list is the
             complete install set rather than only the subset not yet
             installed.
+        skip_runtimepm_marker: Boolean, if True, suppress side effects of computing
+            the install set.
 
     Takes a list of packages of this format:
     {'modalias': 'pci:v000010DEd000010C3sv00003842sd00002670bc03sc03i00',
@@ -1485,7 +1491,12 @@ def already_installed_filter(
 
     # Step 2: Build the installation list with metapackages and modules
     to_install = _build_installation_list(
-        cache, sorted_packages, include_dkms, gpgpu, filter_installed
+        cache,
+        sorted_packages,
+        include_dkms,
+        gpgpu,
+        filter_installed,
+        skip_runtimepm_marker,
     )
 
     # Step 3: Filter out already installed packages
@@ -1504,6 +1515,7 @@ def gpgpu_install_filter(
     get_recommended: bool = True,
     gpgpu: bool = True,
     filter_installed: bool = True,
+    skip_runtimepm_marker: bool = False,
 ) -> List[str]:
     drivers: List[_GpgpuDriver] = []
     allow: List[str] = []
@@ -1524,6 +1536,8 @@ def gpgpu_install_filter(
         filter_installed: Boolean, if False, the returned list is the
             complete install set rather than only the subset not yet
             installed.
+        skip_runtimepm_marker: Boolean, if True, suppress side effects of computing
+            the install set.
 
     Returns:
         A list of drivers to be installed, of the form
@@ -1626,7 +1640,7 @@ def gpgpu_install_filter(
                         # print('Found "recommended" flavour in %s' % (packages[p]))
                 break
     return already_installed_filter(
-        cache, result, include_dkms, gpgpu, filter_installed
+        cache, result, include_dkms, gpgpu, filter_installed, skip_runtimepm_marker
     )
 
 
@@ -1638,6 +1652,7 @@ def auto_install_filter(
     get_recommended: bool = True,
     gpgpu: bool = False,
     filter_installed: bool = True,
+    skip_runtimepm_marker: bool = False,
 ) -> List[str]:
     """
     Get packages which are appropriate for automatic installation.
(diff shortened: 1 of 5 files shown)


debian/changelog
```diff
@@ -1,3 +1,9 @@
+ubuntu-drivers-common (1:0.10.11ubuntu1) UNRELEASED; urgency=medium
+
+  * Add extra_packages to dbus service output
+
+ -- ashton.nelson@canonical.com <ashton.nelson@canonical.com  Wed, 26 Aug 2026 16:51:04 -0500
```
mwhudson: This isn't valid but well we can fix that when we merge with your other branch...

mwhudson (approved):
LGTM, not sure i understand all this code terribly well though

debian/changelog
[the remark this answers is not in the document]
MitchellAugustin: you can drop the changelog entry entirely, I'll need to change it anyway for the actual upload

UbuntuDrivers/service/drivers_service.py
```diff
@@ -45,6 +46,43 @@ def _dbus_error_name(domain: str) -> str:
     return _ERROR_FAILED
 
 
+def _extra_packages(
```
MitchellAugustin: How is this intended to be consumed by users of the dbus service - will service users only be _displaying_ this information and then invoking ubuntu-drivers to handle the installation, or are you trying to gather this information from ubuntu-drivers to then have some other dbus API consumer actually run the `apt install`?

I would definitely prefer having consumers invoke u-d-c to handle the actual installation if possible (to keep things consistent), but in either case, I would prefer not to add another way that the list of available drivers gets computed, on top of what we already have via `system_driver_packages` and `system_gpgpu_driver_packages`, since it seems likely to result in each invoker having a potentially different view of the candidate package list.

Please take a look at how [install_gpgpu](https://github.com/ashuntu/ubuntu-drivers-common/blob/b99ebebbf214dc9279168f56c05753b6377fef59/ubuntu-drivers#L361) and [command_install](https://github.com/ashuntu/ubuntu-drivers-common/blob/b99ebebbf214dc9279168f56c05753b6377fef59/ubuntu-drivers#L259) (for desktop packages) compute their `to_install` lists.
Ideally we should be able to do the same thing here, by preparing the install list via an invocation to system_gpgpu_driver_packages + gpgpu_install_filter for gpgpu, or to get_desktop_package_list for desktop.

If that does not work in your tests, I would like to better understand why, to see if it's something that could be improved in our existing functions, rather than adding another layer of complexity.

MitchellAugustin (changes requested):
Thanks for the contributions. I have some questions about the intended usage of this, as well as some concerns about the duplication of our candidate package list generation logic, inline.

UbuntuDrivers/service/drivers_service.py
ashuntu: The goal is provide an adequate package list and metadata to the App Center so it can make the installs itself. There's a draft PR for the new driver UI here for context: https://github.com/ubuntu/app-center/pull/2170.

Since the App Center is a snap, its only options are either for the dbus service to do installation, or for the App Center to have enough information to install the necessary things over packagekit. Initially we considered this service doing the installation as well, but we want more control from a UI perspective, like users being able to switch their driver branch in a more user-friendly way than `software-properties-gtk` used to provide (which I believe also does some of its own logic to determine what to install, which hopefully we can mostly avoid by providing consistent data via this service).

I think we definitely could do something more in line with what `install_gpgpu` does so we do something more consistent, I'll take a look at it. If you have other ideas or feedback, they're definitely welcome.

MitchellAugustin: @ashuntu Thanks for the added context, that makes sense. In that case, yes, I would just ask that you try to leverage our existing detection functions a bit more directly, as mentioned in my last comment.

UbuntuDrivers/service/drivers_service.py
```diff
@@ -26,6 +26,9 @@
 _ERROR_CACHE_FAILURE = "com.ubuntu.Drivers.Error.CacheFailure"
 _ERROR_FAILED = "com.ubuntu.Drivers.Error.Failed"
 
+# Matches the "ubuntu-drivers install" CLI default.
+_INCLUDE_DKMS = False
```
MitchellAugustin, line 30: +1, I think defaulting to not including DKMS is appropriate since our primary caller of this service will be the Ubuntu driver manager GUI, and since all Ubuntu users who are running supported Ubuntu kernels should be using the prebuilt NVIDIA GPU driver kernel modules anyway.

It _might_ still be nice to expose the option via the dbus interface, even if you don't end up exposing it on the GUI side, just in case some more technical user wants it... but not at all a blocker on my end, and arguably might not be worth the effort, since that's very hypothetical use case that I sort of doubt would end up having any users.

UbuntuDrivers/service/drivers_service.py
```diff
@@ -46,41 +49,59 @@ def _dbus_error_name(domain: str) -> str:
...
+    gpgpu_catalog: Dict[str, PackageInfo],
+) -> Tuple[List[str], List[str]]:
+    """Return ``(packages, gpgpu_packages)`` for *pkg_name*: the install
+    lists that ``ubuntu-drivers install <pkg_name>`` and
+    ``ubuntu-drivers install --gpgpu <pkg_name>`` would each produce.
+    """
+    packages = _install_list(cache, pkg_name, desktop_catalog, gpgpu=False)
+    gpgpu_packages = _install_list(cache, pkg_name, gpgpu_catalog, gpgpu=True)
+
+    if not NvidiaPkgNameInfo(pkg_name).is_valid:
```
MitchellAugustin, line 101: What is this conditional trying to accomplish? As written, I believe this is going to add invalid candidates (those which are lacking a vendor and a flavor) to your returned results, which I think could cause this to deviate from the CLI's resultant list.

UbuntuDrivers/detect.py
```diff
@@ -1665,7 +1692,9 @@ def auto_install_filter(
                 result[p] = packages[p]
         else:
             result[p] = packages[p]
-    return already_installed_filter(cache, result, include_dkms, gpgpu)
+    return already_installed_filter(
+        cache, result, include_dkms, gpgpu, filter_installed
+    )
 
 
 def detect_plugin_packages(
```
MitchellAugustin, line 1724: LGTM, looks like this correctly preserves behavior from other callers who aren't aware of the signature change

tests/test_ubuntu_drivers.py
```diff
@@ -6673,6 +6673,58 @@ def test_gpgpu_install_filter(self):
...
+            kept = UbuntuDrivers.detect.already_installed_filter(
+                cache, pkgs, True, gpgpu=True, filter_installed=False
+            )
+            self.assertEqual(
+                set(kept),
+                set(["nvidia-headless-no-dkms-410", "nvidia-driver-410"]),
+            )
+        finally:
+            chroot.remove()
+
```
MitchellAugustin, line 6745: LGTM

tests/test_drivers_service.py
```diff
@@ -73,6 +81,22 @@ def setup(self, archive):
             capture_output=True,
         )
 
+        for deb_path in installed:
```
MitchellAugustin, line 84: What is the purpose of this addition?

MitchellAugustin (commented):
Thanks @ashuntu , this is much better :)

Please address my inline question around your conditional in _package_lists. After we settle on something for that, I think this will be ready to approve.

UbuntuDrivers/service/drivers_service.py
```diff
@@ -45,6 +49,61 @@ def _dbus_error_name(domain: str) -> str:
...
+            filter_installed=False,
+        )
+
+    return UbuntuDrivers.detect.auto_install_filter(
+        cache,
+        _INCLUDE_DKMS,
+        catalog,
+        pkg_name,
+        get_recommended=False,
+        filter_installed=False,
```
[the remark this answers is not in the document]
MitchellAugustin: @ashuntu this is my preferred resolution to copilot's comment:

> add a flag that suppresses marker creation when generating D-Bus metadata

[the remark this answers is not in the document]
ashuntu: @MitchellAugustin do you think this is a real concern? Not sure of the significance of this file

UbuntuDrivers/service/drivers_service.py
ashuntu: The goal of both of the new fields would be to be a complete set of packages to install for each driver, so I think in certain scenarios the list would not contain anything for non-nvidia drivers, which is why this was added

I think this can be simplified, though

UbuntuDrivers/service/drivers_service.py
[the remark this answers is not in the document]
MitchellAugustin: @ashuntu at minimum, it will be for this feature to be SRUable, since we do not know who might be depending on this file being created and don't want to regress that if it is in use.

UbuntuDrivers/service/drivers_service.py
MitchellAugustin: gotcha. In that case, please just add a code comment documenting that effect.

ashuntu: The way it works now could kind of bypasses the whitelist in `auto_install_filter` so that could be one thing to consider, though I'm not sure if that's useful for a GUI installer, anyway

tests/test_drivers_service.py
ashuntu: I added it to mark certain packages as installed, since we verify that we don't filter against already installed packages

UbuntuDrivers/service/drivers_service.py
ashuntu: One thing I noticed is that the package lists are empty on my 26.10 install, presumably because its in-development and there are no non-dkms options? I suppose that normally wouldn't be a problem for 99% of users

MitchellAugustin: That seems correct. It looks like https://launchpad.net/ubuntu/+source/linux-restricted-signatures-nvidia-7.0 isn't showing anything for stonking yet, so I would expect that it will once we start publishing those.

I would just make sure to test this on Resolute to make sure it works as expected there.

ashuntu: Just tested on Resolute (VM), seems to work as intended. Though I only have NVIDIA devices that need drivers

MitchellAugustin:
@ashuntu in your tests, from your perspective as an end-user, does this seem to execute sufficiently quickly when invoked?

For the sake of SRUability and keeping non-functionality-related code reorganization to a minimum, I am inclined to dismiss copilot's latest concerns if, on a practical system with a standard number of drivers, this all still executes in under a few seconds (since it should only ever be getting invoked from a driver UI that shouldn't need to reload more frequently than that anyway.)

WDYT?

ashuntu:
It executes in about ~4 seconds on my machine. I don't think this is that bad from a UI perspective

MitchellAugustin (approved):
Thanks @ashuntu . I think this is good to go now, approved!

PR #178, opened 2026-09-18
Cherry-pick D-Bus service

ashuntu:


MitchellAugustin (approved):
Thanks @ashuntu , this looks consistent with what we merged into master

PR #179, opened 2026-09-22
detect: prioritize custom GPU config and fix runtimepm/-open handling

KaiChuan-Hsieh:
Respect the custom configuration (custom_supported_gpus.json) when selecting NVIDIA drivers:

* "ubuntu-drivers list" still lists every applicable driver but flags the custom-pinned series as the recommended one.
* "ubuntu-drivers list --recommended" collapses variant driver groups (e.g. the nvidia alternatives) to the single recommended package, while still listing every other suggested package.
* "ubuntu-drivers install" (automatic selection) installs the custom-pinned series instead of the highest-sorted driver. An explicitly requested driver still takes precedence over the custom pin.

Look up custom_supported_gpus.json under /etc and under /usr/share/oem-*-meta, with the file in /etc taking precedence.

Fix runtimepm detection so a candidate driver whose branch carries a variant suffix (e.g. "580-open") or a point release (e.g. "580.1234") is matched against the numeric series, avoiding both the previous exact-match failure and loose substring false-positives.

JIRA: RUISUI-70


KaiChuan-Hsieh:
@MitchellAugustin @mwhudson  

Hello,

Would you please review the MP? to see if it meets your expectation.

Thanks,

UbuntuDrivers/detect.py
```diff
@@ -333,6 +335,21 @@ def apt_cache_modalias_map(
...
+    oem-*-meta package is used instead. If no file exists, the default /etc path
+    is returned.
+    """
+    if os.path.exists(custom_supported_gpus_json):
+        return custom_supported_gpus_json
+
+    for path in sorted(glob.glob(custom_supported_gpus_oem_glob)):
+        if os.path.exists(path):
+            logging.debug("Using custom_supported_gpus.json from %s" % path)
+            return path
```
MitchellAugustin: Is it a guarantee that there is only one oem-*-meta package installed on the system? If not, do we definitely want to pick the first sorted result here?

(I am fine either way, I think this one is the OEM team's decision. Just want to ensure this is what you expect.)

UbuntuDrivers/detect.py
```diff
@@ -764,10 +797,22 @@ def system_driver_packages(
             logging.error(ex)
             return {}
 
+    # Determine which NVIDIA drivers are pinned by the custom configuration
+    # so that they can be prioritized as the recommended candidate.
+    custom_nvidia_packages = set()
+    for alias in modaliases:
+        vid, did = _get_vendor_model_from_alias(alias)
+        if vid == "10DE":
```
MitchellAugustin: Are you sure this is safe to have as case-sensitive?

tests/test_ubuntu_drivers.py
```diff
@@ -2128,13 +2233,29 @@ def test_system_driver_packages_force_install_nvidia(self, mocked_pgcsg):
...
+        # Plain "list" still shows all drivers but prioritizes the custom
+        # configuration as the recommended one.
+        self.assertTrue(res_470_no_390["nvidia-driver-390"]["recommended"])
+        self.assertFalse(res_470_no_390["nvidia-driver-470"]["recommended"])
         packages = UbuntuDrivers.detect.gpgpu_install_filter(
             cache, True, res_470_no_390, "nvidia"
         )
         self.assertEqual(
-            set(packages), set(["linux-modules-nvidia-470-generic-hwe-20.04"])
+            set(packages), set(["linux-modules-nvidia-390-generic-hwe-20.04"])
```
MitchellAugustin: Why does this result change?

MitchellAugustin (commented):
Thanks @KaiChuan-Hsieh - a few questions inline

UbuntuDrivers/detect.py
KaiChuan-Hsieh: As those packages are fully controlled by the OEM team, I think it is okay, as we will allow only one oem folder to carry the custom-supported-gpus.json.

UbuntuDrivers/detect.py
KaiChuan-Hsieh: I am not sure, but I follow the usage in line 409.

tests/test_ubuntu_drivers.py
KaiChuan-Hsieh: It is because of the config in line 2098

KaiChuan-Hsieh:
@MitchellAugustin 

Thanks for your review, please check my comment again.

tests/test_ubuntu_drivers.py
KaiChuan-Hsieh: It is because line 2100 override the preferred driver to 390, so that the 390 driver become recommended.

MitchellAugustin: Gotcha, in that case, we should probably rename that field in the test to `res_390_no_470`

